PR #96, opened 2015-01-13
add service to stop workflow instance

efonsell:


jsyrjala:
Could you squash to one commit? Otherwise looks good.

PR #101, opened 2015-01-16
add service to stop workflow instance

efonsell:


gmokki:
1. Would pause be better term for stop? Meaning that stop sounds like a final act, from which the workflow cannot be recovered.

2. should the sql have extra condition: "and next_activation is not null" so that the stop/pause does not work on already stopped/paused workflow or is it better if the api can be called multiple times

3. I think it would be actually better if next_activation is not touched by the pause/stop operation. For example if someone has scheduled an workflow to activate next monday then pausing/stopping the workflow and then unpausing/resuming it should still honour the original delay.
We do not have a status field on the workflow instance (new,running,stopped,finished), but maybe we could change the group name of the workflow? Or just add the status field to workflow table.

PR #103, opened 2015-01-16
Dynamic state text length

gmokki:


efonsell:
How about using abbreviate instead of left to cut the statetext?

gmokki:
Now uses abbreviate.

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -89,12 +92,39 @@ public void setNamedParameterJdbcTemplate(@NFlow NamedParameterJdbcTemplate nflo
   @Inject
   public void setExecutorDao(ExecutorDao executorDao) {
     this.executorInfo = executorDao;
+    findColumnMaxLengths();
```
efonsell: why findColumnMaxLengths is called in two places?

PR #104, opened 2015-01-16
Add field nflow_instance_action.type.

jsyrjala:
Following types are supported currently:
- stateExecution - for actions created with normal execution. This is also set for old actions created before this feature.
- stateExecutionFailed - for actions that failed due throws exception or retry count was exceeded.
- externalChange - for state changes created manually via API.
- recovery - to indicate workflow was recovered after executor died.


efonsell:
Squash + merge from master/recreate pr

jsyrjala:
New PR #108 with squash and rebase. This one had still bug in alter table statemets.

PR #106, opened 2015-01-16
Add field nflow_workflow_definition.definition_sha1.

jsyrjala:
Field contains SHA1 digest of serialized workflow definition. The field
can be used e.g. quickly detect if definition has changed.


jsyrjala:
Better pull request in #107

PR #119, opened 2015-02-06
update driver

efonsell:


efonsell:
https://jdbc.postgresql.org/documentation/changelog.html#version_9.3-1103

PR #124, opened 2015-02-10
move workflow instance recovery code from executor dao to workflow instance dao

efonsell:


nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/ExecutorDao.java
```diff
@@ -147,25 +145,18 @@ public void updateActiveTimestamp() {
...
-      // TODO: move this block to WorkflowInstanceDao and add transaction
-      int updated = jdbc.update("update nflow_workflow set executor_id = null where id = ? and executor_id in (select id from nflow_executor where " + executorGroupCondition + " and id <> ? and expires < current_timestamp)",
-          instance.id, getExecutorId());
-      if (updated > 0) {
-        WorkflowInstanceAction action = new WorkflowInstanceAction.Builder().setExecutionStart(now()).setExecutionEnd(now())
-            .setExecutorId(getExecutorId()).setType(recovery).setState(instance.state).setStateText("Recovered")
-            .setWorkflowInstanceId(instance.id).build();
-        workflowInstanceDao.insertWorkflowInstanceAction(action);
-      }
+      workflowInstanceDao.recoverWorkflowInstance(instance);
```
gmokki: Should we still create the action in ExexcutorDao? And pass (long instanceId, WorkflowInstanceAction action) as arguments to instanceDao. Then instanceDao would not have to know about the InstanceInfo object at all.

PR #153, opened 2015-07-09
Implement archiving tables

jsyrjala:
Implementing functionality to move old and passive workflows from active table to archive tables.

Adding tables where data from corresponding production table can be moved.
* nflow_archive_workflow
* nflow_archive_workflow_action
* nflow_archive_workflow_state


nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/ArchiveDao.java
```diff
@@ -0,0 +1,110 @@
...
+import com.nitorcreations.nflow.engine.internal.config.NFlow;
+
+@Named
+@DependsOn(NFLOW_DATABASE_INITIALIZER)
+public class ArchiveDao {
+  private JdbcTemplate jdbc;
+  private TableMetadataChecker tableMetadataChecker;
+
+  @Inject
+  public void setJdbcTemplate(@NFlow JdbcTemplate jdbcTemplate) {
```
efonsell: Could we use field injection here?

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/TableMetadataChecker.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    public String toString() {
+      return ReflectionToStringBuilder.toString(this, SHORT_PREFIX_STYLE);
+    }
+  }
+
+  @Inject
+  public void setJdbcTemplate(@NFlow JdbcTemplate jdbcTemplate) {
```
efonsell: use field injection?

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/TableMetadataChecker.java
```diff
@@ -0,0 +1,113 @@
...
+      }
+    }
+  }
+
+  private Map<String, ColumnMetadata> getMetadata(String tableName) {
+    return jdbc.query("select * from " + tableName + " where 1 = 0", new MetadataExtractor());
+  }
+
+  private static class MetadataExtractor implements ResultSetExtractor<Map<String, ColumnMetadata>> {
+    private Map<String, String> typeAliases = typeAliases();
```
efonsell: this could be static map

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -262,8 +264,9 @@ protected void doInTransactionWithoutResult(TransactionStatus status) {
         updateWorkflowInstance(instance);
         int parentActionId = insertWorkflowInstanceAction(instance, action);
         for (WorkflowInstance childTemplate : childWorkflows) {
-          WorkflowInstance childWorkflow = new WorkflowInstance.Builder(childTemplate).setParentWorkflowId(instance.id)
-              .setParentActionId(parentActionId).build();
+          Integer rootWorkflowId = instance.rootWorkflowId != null ? instance.rootWorkflowId : instance.id;
```
efonsell: prefer == instead of !=: instance.rootWorkflowId == null ? instance.id : instance.rootWorkflowId

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/workflow/instance/WorkflowInstance.java
```diff
@@ -43,6 +43,12 @@
   public final Integer executorId;
 
   /**
+   * The id of the workflow that created the hierarchy of workflow where this sub workflow belongs to.
+   * Null for workflows without children and workflows that are root of hierarchy.
```
efonsell: Why it is null for workflows without children and root workflows?

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/ArchiveDao.java
jsyrjala: JdbcTemplate is injected via a setter elsewhere. And I recall that there was some reason for that?

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/workflow/instance/WorkflowInstance.java
efonsell: Shouldn't it be "Null for workflows that are the root of the hierarchy"? Child workflows may not have their own children but they may belong to a hierarchy so they should have root id set.

jsyrjala: Yes.

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/workflow/instance/WorkflowInstance.java
```diff
@@ -43,6 +43,12 @@
...
+   */
+  public final Integer rootWorkflowId;
+
+  /**
    * The id of the workflow that created this sub workflow. Is null for parent workflows.
```
jsyrjala: Should be "Is null for root workflows"

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/ArchiveDao.java
```diff
@@ -0,0 +1,114 @@
...
+
+  private void archiveStateTable(String workflowIdParams) {
+    String columns = columnsFromMetadata("nflow_workflow_state");
+    jdbc.update("insert into nflow_archive_workflow_state (" + columns + ") " +
+            "select " + columns + " from nflow_workflow_state where workflow_id in " + workflowIdParams);
+  }
+
+  private void deleteWorkflows(String workflowIdParams) {
+    jdbc.update("delete from nflow_workflow_state where workflow_id in " + workflowIdParams);
+    jdbc.update("update nflow_workflow set root_workflow_id=null, parent_workflow_id=null, parent_action_id=null " +
```
gmokki: do we need to do the update? can we reorder the deletes instead?

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/ArchiveDao.java
```diff
@@ -0,0 +1,114 @@
...
+            "where id in " + workflowIdParams);
+    jdbc.update("delete from nflow_workflow_action where workflow_id in " + workflowIdParams);
+    jdbc.update("delete from nflow_workflow where id in " + workflowIdParams);
+  }
+
+  private String columnsFromMetadata(String tableName) {
+    List<String> columnNames = jdbc.query("select * from " + tableName + " where 1 = 0", columnNamesExtractor);
+    return join(columnNames.toArray(), ",");
```
gmokki: no need for toArray conversion here? StringUtils join supports collections directly.

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/ArchiveDao.java
```diff
@@ -0,0 +1,114 @@
...
+                    "  and parent.root_workflow_id is null " +
+                    "  and not exists(" +
+                    "    select 1 from nflow_workflow child where child.root_workflow_id = parent.id " +
+                    "      and (child.modified > ? or child.next_activation is not null)" +
+                    "  )" +
+                    "  order by modified asc " +
+                    "  limit " + maxRows +
+                    ") as archivable_parent " +
+                    "where archivable_parent.id = w.id or archivable_parent.id = w.root_workflow_id",
+            new Object[] { toTimestamp(before), toTimestamp(before) }, new ArchivableWorkflowsRowMapper());
```
gmokki: swap arguments, rowmapper before arguments -> can get rid of the new Object[] (java compiler adds it automatically)

nflow-engine/src/main/resources/scripts/db/mysql.create.ddl.sql
```diff
@@ -64,3 +71,57 @@ create table if not exists nflow_workflow_definition (
...
+  executor_id int,
+  retries int not null default 0,
+  created timestamp(3) not null,
+  modified timestamp(3) not null,
+  executor_group varchar(64) not null,
+  constraint nflow_archive_workflow_uniq unique (type, external_id, executor_group)
+);
+
+drop index nflow_archive_workflow_activation;
+create index nflow_archive_workflow_activation on nflow_archive_workflow(next_activation, modified);
```
gmokki: no need to create index with next_activation to archive tables

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/ArchiveDao.java
jsyrjala: nflow_workflow has circular foreign key references (parent_workflow_id points to another nflow_workflow row). Updates break this cycle. I dont see any easy way to reorder deletes to achieve this.

gmokki: True. so the only way to do it would be to sort the workflow_ids we operate on from root to leaf in the hierarchy. And insert to archive tables from root to leaf and delete in reverse order from leaf to root?

But on the other hand we have on delete cascades on the foreign keys so we actually only need to delete rows where nflow_parent_id is not null and the cascade should handle the rest. right?

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/service/ArchiveService.java
```diff
@@ -0,0 +1,58 @@
...
+    do {
+      workflowIds = archiveDao.listArchivableWorkflows(olderThan, batchSize);
+      if (workflowIds.isEmpty()) {
+        break;
+      }
+      archiveDao.archiveWorkflows(workflowIds);
+      archivedWorkflows += workflowIds.size();
+
+      double timeDiff = stopWatch.getTime() / 1000.0;
+      log.debug("Archived {} workflows. {} workflows / second. Workflow ids: {}. ", workflowIds.size(), archivedWorkflows
```
eputtone: It would be nice, if archiving could say "hello" periodically (info-level). For example, it could report the total number of deleted (root) workflow instances in the next iteration after 1 minute has elapsed from previous "hello". It would strengthen the trust things are going forward during long archiving process without the need to monitor database.

eputtone: Actually ArchiveDao.archiveWorkflowTable() could return the number of workflow instances deleted... so "hello" would contain the number of any type of workflow instance.

efonsell:
Deprecated by archiving2

PR #159, opened 2015-10-30
Implement archiving tables

efonsell:
Implementing functionality to move old and passive workflows from active table to archive tables.

Adding tables where data from corresponding production table can be moved.
* nflow_archive_workflow
* nflow_archive_workflow_action
* nflow_archive_workflow_state


nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/util/PeriodicLogger.java
```diff
@@ -0,0 +1,34 @@
...
+    public void log(String message, Object ... parameters) {
+        long now = periodNumber();
+        if(previousLogging == null || previousLogging != now) {
+            logger.info(message, parameters);
+        }
+        previousLogging = now;
+    }
+
+    private long periodNumber() {
+        return DateTime.now().getMillis() / 1000 / periodInSeconds;
```
gmokki: currentTimeMillis() instead of creating a new DateTime() object

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/service/ArchiveService.java
```diff
@@ -0,0 +1,64 @@
...
+      }
+      archivedWorkflows = archiveDao.archiveWorkflows(workflowIds);
+      archivedWorkflowsTotal += archivedWorkflows;
+
+      double timeDiff = max(stopWatch.getTime() / 1000.0, 0.000001);
+      log.debug("Archived {} workflows. {} workflows / second. Workflow ids: {}. ", archivedWorkflows, archivedWorkflowsTotal
+          / timeDiff, workflowIds);
+      periodicLogger.log("Archived {} workflows. Archiving about {} workflows / second.", archivedWorkflows,
+          archivedWorkflowsTotal / timeDiff);
+    } while (!workflowIds.isEmpty());
```
gmokki: can we ever hit this condition - I think this should be a while (true) loop instead

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/DaoUtil.java
```diff
@@ -35,4 +38,21 @@ public static Integer getInt(ResultSet rs, String columnLabel) throws SQLExcepti
...
+
+  public static final class ColumnNamesExtractor implements ResultSetExtractor<List<String>> {
+    static final ColumnNamesExtractor columnNamesExtractor = new ColumnNamesExtractor();
+
+    private ColumnNamesExtractor() {
+    }
+
+    @Override
+    public List<String> extractData(ResultSet rs) throws SQLException, DataAccessException {
+      List<String> columnNames = new LinkedList<>();
```
gmokki: arraylist instead? we're not going to remove column names from the middle of the list?

PR #160, opened 2015-11-06
update deps and plugins

efonsell:


jsyrjala:
Please, squash these commits to a single one.

PR #168, opened 2016-02-06
set workflow instance next activation to now by default

efonsell:


eputtone:
Shouldn't this be handled by setting default value of nflow_workflow.next_activation to current timestamp instead of handling it in Java? We are trying to move towards solution, where all timing is based on database clock instead of the clocks of hosts running nflow engine.

    next_activation timestamp(3) null default current_timestamp

PR #169, opened 2016-02-06
update swagger-ui, cxf, hikaricp and jackson

efonsell:


nflow-jetty/src/main/resources/static/doc/index.html
```diff
@@ -18,6 +19,7 @@
   <script src='lib/backbone-min.js' type='text/javascript'></script>
   <script src='swagger-ui.js' type='text/javascript'></script>
   <script src='lib/highlight.7.3.pack.js' type='text/javascript'></script>
+  <script src='lib/jsoneditor.min.js' type='text/javascript'></script>
```
eputtone: jsoneditor.min.js needs to be above swagger-ui.js -dependency. Currently swagger-ui does not find JSONEditor and Swagger UI fails to open.

PR #179, opened 2016-03-25
tune logging

efonsell:


nflow-rest-api/src/main/java/com/nitorcreations/nflow/rest/v1/converter/ListWorkflowInstanceConverter.java
```diff
@@ -83,8 +82,8 @@ private JsonNode stringToJson(String key, String value) {
     try {
       return nflowRestObjectMapper.readTree(value);
     } catch (IOException e) {
-      logger.warn("Failed to parse state variable {} value as JSON, returning value as unparsed string: {}: {}",
-    		  key, e.getClass().getSimpleName(), e.getMessage());
+      logger.debug("Failed to parse state variable {} value as JSON, returning value as unparsed string: {}: {}", key,
```
jsyrjala: Is this change needed? Something is pretty wrong if state variable is not a json value.

jsyrjala: I guess it should noisily log if the value is expected to be json and it is not.

Maybe something like this.
````
// check if the first character is suitable for json (object, array, string, number, boolean or null
if (value.matches("\s*[\{\[\"01-9tfn]”)) { 
   // try to parse with nflowRestObjectMapper (log warn if failure)
} else {
   // this doesn't look like JSON at all. Return as a string.
}

efonsell: Yes it is needed, altough the whole json conversion is controversial. State variable values are not necessarily always json.

PR #184, opened 2016-03-26
log warnings when state processor threads are stuck

efonsell:


nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/executor/WorkflowDispatcher.java
```diff
@@ -55,6 +55,9 @@ public void run() {
 
           if (!shutdownRequested) {
             executorRecovery.tick();
+            if (stateProcessorFactory.getPotentiallyStuckProcessors() == executor.getThreadCount()) {
+              logger.warn("All state processor threads are potentially stuck.");
```
jsyrjala: Once things get stuck, do we need the logging at every polling interval? 

Maybe use this https://github.com/NitorCreations/nflow/blob/master/nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/util/PeriodicLogger.java

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/executor/WorkflowStateProcessorFactory.java
```diff
@@ -38,7 +51,23 @@ public WorkflowStateProcessorFactory(WorkflowDefinitionService workflowDefinitio
...
+        workflowInstancePreProcessor, env, processingInstances, listeners);
+  }
+
+  public int getPotentiallyStuckProcessors() {
+    DateTime now = now();
+    int potentiallyStuck = 0;
+    synchronized (processingInstances) {
+      for (Entry<Integer, DateTime> entry : processingInstances.entrySet()) {
+        int processingTimeMinutes = minutesBetween(entry.getValue(), now).getMinutes();
+        if (processingTimeMinutes > 5) {
```
jsyrjala: Make value configurable.

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/executor/WorkflowStateProcessorFactory.java
```diff
@@ -23,6 +34,8 @@
   private final Environment env;
   @Autowired(required = false)
   protected WorkflowExecutorListener[] listeners = new WorkflowExecutorListener[0];
+  final Map<Integer, DateTime> processingInstances = synchronizedMap(new HashMap<Integer, DateTime>());
```
jsyrjala: LinkedHashMap

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/executor/WorkflowStateProcessorFactory.java
jsyrjala: And I think that the unit should be in seconds.

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/executor/WorkflowStateProcessorFactory.java
gmokki: Why LinkedHashMap? Nothing in the code relies on the insertion order and triple-memory usage of LinkedHashMap with much slower operations is unneeded overhead.
I propose we just use ConcurrentHashMap -> no need to synchronize even in getPotentiallyStuckProcessors()

jsyrjala: With linkedhashmap order where stuck threads are logged ( https://github.com/NitorCreations/nflow/pull/184/commits/f7b3e030fdb0f72a946386d18ea4a03132f510b5#diff-ec7adb4d50d2123d1c7a446126265b75R65 ) would stay the same. But ConcurrentHashMap would be better.

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/executor/WorkflowDispatcher.java
gmokki: Good idea. I just added a pull request to add support for logging warnings through PeriodicLogger
https://github.com/NitorCreations/nflow/pull/185

jsyrjala: I merged #185 to master. 

Also, would it be interesting to print stuck threads count also when everything is not stuck.
Something like this:

       7 of 10 state processor threads are potentially stuck (processing more than N seconds).

efonsell: Done

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/executor/WorkflowStateProcessorFactory.java
efonsell: Done

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/executor/WorkflowStateProcessorFactory.java
efonsell: Done

PR #188, opened 2016-04-01
do not rollback transaction when inserting workflow instance that already exists

efonsell:


nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -167,27 +171,32 @@ private int insertWorkflowInstanceWithTransaction(final WorkflowInstance instanc
...
+              ps.setString(p++, abbreviate(instance.stateText, instanceStateTextLength));
+              ps.setTimestamp(p++, toTimestamp(instance.nextActivation));
+              return ps;
+            }
+          }, keyHolder);
+        } catch (DuplicateKeyException e) {
+          logger.warn("Failed to insert workflow instance", e);
+          return -1;
+        }
         int id = keyHolder.getKey().intValue();
```
jsyrjala: If it gets DuplicateKeyException above, this will insert variables again? => there will be double set of variables?

Is the end result same as with insertWorkflowInstanceWithCte() ?

Would it be better to not modify variables in case of duplicate key exception?

nflow-engine/src/test/java/com/nitorcreations/nflow/engine/internal/dao/WorkflowInstanceDaoTest.java
```diff
@@ -183,6 +183,35 @@ public void updateWorkflowInstanceWithRootWorkflowAndChildWorkflowsWorks() {
   }
 
   @Test
+  public void updateWorkflowInstanceWithChildWorkflowHavingExistinExternalIdWorks() {
```
jsyrjala: Tests should try to see what happens if there are variables.

nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/WorkflowInstanceDao.java
efonsell: No, it returns -1 -> variable will not be inserted.
Yes, same end result.
They are not modified in case of duplicate key exception.

nflow-engine/src/test/java/com/nitorcreations/nflow/engine/internal/dao/WorkflowInstanceDaoTest.java
efonsell: Fixed

PR #192, opened 2016-04-12
add abstract base class for model objects

efonsell:


eputtone:
Left out on purpose?
- com.nitorcreations.nflow.engine.workflow.executor.WorkflowExecutor
- com.nitorcreations.nflow.engine.internal.workflow.StateExecutionImpl

efonsell:
@eputtone WorkflowExecutor and StateExecutionImpl inclulded

jsyrjala:
squash commit: fix after merge			81d3e92

nflow-engine/src/main/java/io/nflow/engine/model/ModelObject.java
```diff
@@ -0,0 +1,25 @@
+package io.nflow.engine.model;
+
+import static org.apache.commons.lang3.builder.EqualsBuilder.reflectionEquals;
+import static org.apache.commons.lang3.builder.HashCodeBuilder.reflectionHashCode;
+import static org.apache.commons.lang3.builder.ToStringBuilder.reflectionToString;
+import static org.apache.commons.lang3.builder.ToStringStyle.SHORT_PREFIX_STYLE;
+
+public abstract class ModelObject {
```
jsyrjala: Add javadoc to describe following:
- inheriting ModelObject is an implementation detail only to provide toString/equals/hashCode for subclasses. We would use something more sensible method, but this is what Java allows.
- user code should never user ModelObject for anything

PR #193, opened 2016-04-12
Fb fixes

efonsell:


nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/storage/db/OracleDatabaseConfiguration.java
```diff
@@ -26,7 +27,7 @@
 
   private static final Logger logger = getLogger(OracleDatabaseConfiguration.class);
   public static final String DB_TYPE_ORACLE = "oracle";
-  static boolean useBatchUpdates;
+  private boolean useBatchUpdate;
```
jsyrjala: This does not match commit comment.

PR #196, opened 2016-04-23
set active and expires when inserting row to nflow_executor table

efonsell:


jsyrjala:
Update/migration script missing for database that adds not nulls.

PR #197, opened 2016-04-23
wake dispatcher when new workflow instance is created

efonsell:


nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/executor/ThresholdBlockingQueue.java
```diff
@@ -20,7 +20,7 @@ public ThresholdBlockingQueue(int capacity, int notifyThreshHold) {
     queue = new LinkedBlockingQueue<>(capacity);
   }
 
-  private void notifyIf() {
+  public void notifyIf() {
```
jsyrjala: Why is this called notifyIf? Better name, please.

PR #203, opened 2016-06-04
wakeup parent in expected states only

efonsell:


nflow-engine/src/main/java/com/nitorcreations/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -405,21 +405,25 @@ public boolean updateNotRunningWorkflowInstance(WorkflowInstance instance) {
...
         + "least(current_timestamp, coalesce(next_activation, current_timestamp)) else next_activation end), "
         + "external_next_activation = current_timestamp where " + executorInfo.getExecutorGroupCondition()
-        + " and id = ? and next_activation is not null";
-    return jdbc.update(sql, workflowInstanceId) == 1;
+        + " and id = ? and next_activation is not null");
+    return addExpectedStatesToQueryAndUpdate(sql, workflowInstanceId, expectedStates);
   }
 
-  public boolean wakeupWorkflowInstanceIfNotExecuting(long id, String[] expectedStates) {
+  public boolean wakeupWorkflowInstanceIfNotExecuting(long workflowInstanceId, String[] expectedStates) {
```
eputtone: This method seems like dead code to me. Only called by tests and WorkflowService.wakeupWorkflowInstance(), which in turn is used from nowhere. Simplify and remove the method and places it is used from?

eputtone: Forget this, WorkflowService.wakeupWorkflowInstance() is useful for users of nFlow.

PR #206, opened 2016-07-15
Findbugs contrib

efonsell:


gmokki:
Looks good. Everything I was about to complain was upon second inspection already handled better than I had planned to propose.

nflow-engine/src/main/java/io/nflow/engine/internal/executor/WorkflowStateProcessor.java
```diff
@@ -55,7 +56,7 @@
   private static final PeriodicLogger threadStuckLogger = new PeriodicLogger(logger, 60);
   private static final String MDC_KEY = "workflowInstanceId";
 
-  private final int MAX_SUBSEQUENT_STATE_EXECUTIONS = 100;
+  private static final int MAX_SUBSEQUENT_STATE_EXECUTIONS = 100;
```
eputtone: Perhaps it would be nice, if this delimiter would be externally configurable. The correct value is derived from your business case (maybe you want to execute something hundreds of times quickly) and your will to guard against programming errors (maybe you dislike potential busy loop in sending emails for example).

Add a task to Trello if agree?

jsyrjala:
Squash some of the commits.

nflow-engine/src/main/java/io/nflow/engine/internal/executor/WorkflowStateProcessor.java
efonsell: Added to Trello

PR #209, opened 2016-11-04
add putStateVariable that takes optional value

efonsell:


nflow-engine/src/main/java/io/nflow/engine/workflow/instance/WorkflowInstance.java
```diff
@@ -421,11 +423,29 @@ public Builder putStateVariable(String key, Object value) {
...
 
     /**
+     * Put a state variable to the state variables map if the optional value is present.
+     * @param key The name of the variable.
+     * @param value The optional value of the variable, serialized by object mapper.
+     * @return this.
+     */
+    @SuppressFBWarnings(value = "WEM_WEAK_EXCEPTION_MESSAGING", justification = "exception message is ok")
+    public Builder putStateVariable(String key, Optional<?> optionalValue) {
+      return optionalValue.map(value -> putStateVariable(key, value)).orElse(this);
```
efonsell: Latter. Javadoc clarified.

PR #219, opened 2017-04-02
add started timestamp to nflow_workflow table

efonsell:


eputtone:
Add SQL script for initializing nflow_workflow.started from nflow_workflow_action table to migration SQL scripts?

PR #221, opened 2017-04-09
set state variables when creating or updating workflow instance via REST API

efonsell:


eputtone:
I can't find an integration test that checks that updating state variables actually works?

PR #222, opened 2017-04-19
Fix #212: make jaxrs server address configurable

### LINKED ISSUE #212: Change NflowJettyConfiguration address to be configurable
Hi,

Could following line in NflowJettyConfiguration class

  factory.setAddress('/' + factory.getAddress());

be changed to be configurable like:

  factory.setAddress(env.getProperty("nflow.api.url"));

//Juha

efonsell:


nflow-jetty/src/main/java/io/nflow/jetty/config/NflowJettyConfiguration.java
```diff
@@ -69,7 +69,7 @@ public Server jaxRsServer(WorkflowInstanceResource workflowInstanceResource,
         statisticsResource,
         archiveResource
         ));
-    factory.setAddress('/' + factory.getAddress());
+    factory.setAddress(env.getProperty("nflow.api.basepath", '/' + factory.getAddress()));
```
jsyrjala: When `nflow.api.basepath` is not defined, does this work identically as before?

PR #229, opened 2017-08-15
Remove @PostProcess that causes database access before start()

ttiurani:
Related:

https://github.com/NitorCreations/nflow/pull/228

### DIFF
--- a/nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
+++ b/nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
@@ -39,7 +39,6 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 
-import javax.annotation.PostConstruct;
 import javax.inject.Inject;
 
 import org.joda.time.DateTime;
@@ -99,8 +98,8 @@ public class WorkflowInstanceDao {
   private long workflowInstanceQueryMaxResultsDefault;
   private long workflowInstanceQueryMaxActions;
   private long workflowInstanceQueryMaxActionsDefault;
-  int instanceStateTextLength;
-  int actionStateTextLength;
+  int instanceStateTextLength = -1;
+  int actionStateTextLength = -1;
 
   @Inject
   public void setSqlVariants(SQLVariants sqlVariants) {
@@ -147,10 +146,18 @@ public class WorkflowInstanceDao {
     this.workflowInstanceFactory = workflowInstanceFactory;
   }
 
-  @PostConstruct
-  public void findColumnMaxLengths() {
-    instanceStateTextLength = jdbc.query("select state_text from nflow_workflow where 1 = 0", firstColumnLengthExtractor);
-    actionStateTextLength = jdbc.query("select state_text from nflow_workflow_action where 1 = 0", firstColumnLengthExtractor);
+  private int getInstanceStateTextLength() {
+    if (instanceStateTextLength == -1) {
+      instanceStateTextLength = jdbc.query("select state_text from nflow_workflow where 1 = 0", firstColumnLengthExtractor);
+    }
+    return instanceStateTextLength;
+  }
+
+  int getActionStateTextLength() {
+    if (actionStateTextLength == -1) {
+      actionStateTextLength = jdbc.query("select state_text from nflow_workflow_action where 1 = 0", firstColumnLengthExtractor);
+    }
+    return actionStateTextLength;
   }
 
   public int insertWorkflowInstance(WorkflowInstance instance) {
@@ -172,7 +179,7 @@ public class WorkflowInstanceDao {
       sqlb.append("with wf as (").append(insertWorkflowInstanceSql()).append(" returning id)");
       Object[] instanceValues = new Object[] { instance.type, instance.rootWorkflowId, instance.parentWorkflowId,
           instance.parentActionId, instance.businessKey, instance.externalId, executorInfo.getExecutorGroup(),
-          instance.status.name(), instance.state, abbreviate(instance.stateText, instanceStateTextLength),
+          instance.status.name(), instance.state, abbreviate(instance.stateText, getInstanceStateTextLength()),
           toTimestamp(instance.nextActivation), instance.signal.orElse(null) };
       int pos = instanceValues.length;
       Object[] args = Arrays.copyOf(instanceValues, pos + instance.stateVariables.size() * 2);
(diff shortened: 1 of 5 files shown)


nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -99,8 +98,8 @@
   private long workflowInstanceQueryMaxResultsDefault;
   private long workflowInstanceQueryMaxActions;
   private long workflowInstanceQueryMaxActionsDefault;
-  int instanceStateTextLength;
-  int actionStateTextLength;
+  int instanceStateTextLength = -1;
```
jsyrjala: Any reason why these are not private?

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -147,10 +146,18 @@ public void setWorkflowInstanceFactory(WorkflowInstanceFactory workflowInstanceF
...
-    instanceStateTextLength = jdbc.query("select state_text from nflow_workflow where 1 = 0", firstColumnLengthExtractor);
-    actionStateTextLength = jdbc.query("select state_text from nflow_workflow_action where 1 = 0", firstColumnLengthExtractor);
+  private int getInstanceStateTextLength() {
+    if (instanceStateTextLength == -1) {
+      instanceStateTextLength = jdbc.query("select state_text from nflow_workflow where 1 = 0", firstColumnLengthExtractor);
+    }
+    return instanceStateTextLength;
+  }
+
+  int getActionStateTextLength() {
```
jsyrjala: Missing private modifier?

nflow-tests/src/test/java/io/nflow/tests/SkipAutoStartTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import io.nflow.tests.runner.NflowServerRule;
+
+public class SkipAutoStartTest extends AbstractNflowTest {
+
+  @ClassRule
+  public static NflowServerRule server = new NflowServerRule.Builder()
```
jsyrjala: Would be better to have some comment/explicit code that configures non working database connection.

nflow-engine/src/main/java/io/nflow/engine/internal/executor/WorkflowDispatcher.java
```diff
@@ -80,6 +86,8 @@ public void run() {
           sleep(false);
         }
       }
+    } catch (IOException | ReflectiveOperationException e) {
```
jsyrjala: Either move the try-catch-logging-rethow around `workflowDefinitions.postProcessWorkflowDefinitions()` or remove catch statement altogether.

Currently if normal operation should throw IOExceptions this would catch and hide them.

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
efonsell: No, it needs to be package private to avoid synthetic accessors

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
ttiurani: No other than those didn't have private before, so didn't think to check. Added privates in upcoming commit.

nflow-tests/src/test/java/io/nflow/tests/SkipAutoStartTest.java
ttiurani: Having a non-database connection didn't work, as there is code that establishes the connection to the database here:

https://github.com/NitorCreations/nflow/blob/master/nflow-engine/src/main/java/io/nflow/engine/internal/storage/db/DatabaseConfiguration.java#L36

That does (and IMO should) fail when initializing NFlow. But I can add comments about this to the test case.

nflow-engine/src/main/java/io/nflow/engine/internal/executor/WorkflowDispatcher.java
ttiurani: True, my bad. How about catching (Exception e) there? I can't remove the catch as I can't add a throws to the run() method, and adding a surrounding try catch around the definition search is kind of silly, as I do want to shutdown stuff in the finally block after that and not go into the loop.

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
ttiurani: No, sorry WorkflowInstanceDaoTest.java sets them manually, so that's why they are package private. I'll not change that, if that's ok.

nflow-engine/src/main/java/io/nflow/engine/internal/executor/WorkflowDispatcher.java
jsyrjala: rethrow exception, then the loop is skipped?

```
try {
  workflowDefinitions.postProcessWorkflowDefinitions();
} catch (Exception e) {
  logger.error("Fetching definitions failed");
  throw e;
}
```

jsyrjala: Works also if the catch is removed completely.

ttiurani: Hmm, I still don't quite understand: the above code does not work, as that causes compilation error in the run() method which doesn't throw anything. Same if I remove the catch.

This:

```
    } catch (Exception e) {
      logger.error("Fetching workflow definitions failed", e);
    } finally {
      shutdownPool();
      ....
```

works, but can't wrap my head around how your suggestions do?

jsyrjala: Never mind, I took another look at nesting, your original code is correct.

ttiurani: So my two commits just now added your better exception description, but otherwise the logic is the same.

nflow-tests/src/test/java/io/nflow/tests/SkipAutoStartTest.java
ttiurani: So now the test case has comments.

PR #230, opened 2017-08-24
Add retry to workflow instance state saving

eputtone:


nflow-engine/src/main/java/io/nflow/engine/internal/executor/WorkflowStateProcessor.java
```diff
@@ -205,6 +209,26 @@ private WorkflowInstance saveWorkflowInstanceState(StateExecutionImpl execution,
...
+        try {
+          Thread.sleep(SECONDS.toMillis(stateSaveRetryDelay));
+        } catch (@SuppressWarnings("unused") InterruptedException ok) {
+        }
+      }
+    } while (stateSaveRetryEnabled);
+    throw new IllegalStateException("Failed to save workflow instance new state");
+  }
+
+  public void setStateSaveRetryEnabled(boolean stateSaveRetryEnabled) {
```
jsyrjala: Add comment that this is intended only for automated tests?

eputtone: Fixed

PR #231, opened 2017-10-27
Properties for explicit DB column max lengths

eputtone:
In one deployment (with Java 7 and nFlow 3.3.0), H2 strangely returned 0 as column length:
- https://github.com/NitorCreations/nflow/blob/4.0.0/nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java#L139
- https://github.com/NitorCreations/nflow/blob/4.0.0/nflow-engine/src/main/java/io/nflow/engine/internal/dao/DaoUtil.java#L33

As a result, state transition failed (StringUtils.abbreviate() error "minimum abbreviation width 4").

The problem could not be reproduced in test deployment. This PR enables setting column lengths through properties, so that such problems can be at least evaded, if ever encountered again.

### DIFF
--- a/nflow-engine/src/main/java/io/nflow/engine/internal/dao/ExecutorDao.java
+++ b/nflow-engine/src/main/java/io/nflow/engine/internal/dao/ExecutorDao.java
@@ -54,7 +54,7 @@ public class ExecutorDao {
   String executorGroupCondition;
   int timeoutSeconds;
   int executorId = -1;
-  int hostMaxLength = -1;
+  int hostMaxLength;
 
   @Inject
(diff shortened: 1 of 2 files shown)


gmokki:
Is it possible that the failing case was run with very old H2 library version? If that could be verified we could just refuse to run with too old H2?

eputtone:
It is possible, but would require more investigation (+source codes) to say for certain

For instance, the deployment environment has JBoss and nFlow dependencies are defined using maven-dependency-plugin (?), not as Maven dependencies.

So there's more moving parts than in many other nFlow deployments.

PR #270, opened 2018-11-18
Automatic refresh for workflow instance page in Explorer

eputtone:
- Refresh interval determined by existing configuration `radiator.pollPeriod`
- Includes small related improvement: unregister workflow definition poller when navigating away from definition view instead of creating a new poller for each workflow definition view visited


nflow-explorer/src/app/workflow/workflow.js
```diff
@@ -4,15 +4,29 @@
...
+      parentWorkflow, childWorkflows, $scope, config, $interval, WorkflowService) {
     var self = this;
     self.workflow = workflow;
     self.parentWorkflow = parentWorkflow;
     self.definition = definition;
     self.childWorkflows = childWorkflows;
+
+    function reloadWorkflow(workflowId) {
+      console.log('Fetching workflow id ' + workflowId);
+      WorkflowService.get(workflowId).then(function(workflow) {
```
gmokki: If there is already a pending fetch - should we cancel it or skip this fetch? This is for example when the network is down/slow.

nflow-explorer/src/app/workflow/workflow.js
```diff
@@ -4,15 +4,29 @@
...
+        self.workflow = workflow;
+      });
+    }
+    self.poller = $interval(function() { reloadWorkflow(self.workflow.id); },
+      config.radiator.pollPeriod * 1000);
+    $scope.$on('$destroy', function() {
```
gmokki: should we stop the poller if the workflow is in final state?

eputtone: No, because someone can move it manually to some non-final state? If someone wants to monitor finished workflow, I think we can grant him 1 rest api call every 15 seconds without overloading the servers :)

nflow-explorer/src/app/workflow/workflow.js
eputtone: Good point, though the previous pollers (executors, workflow definition) don't do this either. Maybe we should add a task for fixing all the pollers?

PR #276, opened 2018-12-20
Popup notifications from REST API calls

eputtone:


efonsell:
When nFlow REST API is not running at all, I get "Aborted by browser, check network connection and cors" error.

eputtone:
> When nFlow REST API is not running at all, I get "Aborted by browser, check network connection and cors" error.

Yep, $http gives status -1 to the interceptor when browser aborts the request due to connection timeout, which is what happens here.
```
Any response status code outside of that range is considered an error status and will result in the error callback being called. Also, status codes less than -1 are normalized to zero. -1 usually means the request was aborted, e.g. using a config.timeout.
```
https://docs.angularjs.org/api/ng/service/$http#$http-returns

What would a better error message?

PR #286, opened 2019-02-12
Bulk worklfow demo + test

gmokki:


nflow-tests/src/main/java/io/nflow/tests/demo/workflow/BulkWorkflow.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+    long progress = completed * 100 / childWorkflows.size();
+    return retryAfter(now().plusMinutes(15), "Waiting for child workflows to complete - " + progress + "% done");
+  }
+
+  private void wakeup(WorkflowInstance instance) {
+    instanceService.wakeupWorkflowInstance(instance.id, emptyList());
+  }
+
+  private boolean isRunning(WorkflowInstance instance) {
```
gmokki: This should be `protected` so that extending class can override it - sometimes you want to start another workflow if the main flow goes to manual step for example.

nflow-tests/src/test/java/io/nflow/tests/BulkWorkflowTest.java
```diff
@@ -0,0 +1,55 @@
...
+    req.stateVariables.put("requestData", nflowObjectMapper().valueToTree(asList(1, 2, 3, 4, 5, 6, 6, 7, 8, 9, 10)));
+    CreateWorkflowInstanceResponse resp = fromClient(workflowInstanceResource, true).put(req,
+        CreateWorkflowInstanceResponse.class);
+    assertThat(resp.id, notNullValue());
+    workflowId = resp.id;
+  }
+
+  @Test(timeout = 30000)
+  public void t02_waitForBulkToFinish() throws InterruptedException {
+    getWorkflowInstance(workflowId, done.name());
```
gmokki: We should actually test something like that the start time of the first and last child workflow should be at least few seconds apart.

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/BulkWorkflow.java
```diff
@@ -0,0 +1,137 @@
...
+          "No child workflows found - either add them before starting the parent or implement splitWorkflowImpl");
+    }
+    return true;
+  }
+
+  public NextAction waitForChildrenToFinish(StateExecution execution,
+      @StateVar(value = "concurrency", readOnly = true) int concurrency) {
+    List<WorkflowInstance> childWorkflows = execution.getAllChildWorkflows();
+    long running = childWorkflows.stream().filter(this::isRunning).count();
+    long completed = childWorkflows.stream().filter(this::isFinished).count();
```
gmokki: This will now loop through the upto 100k size children twice. We could do it with streams by categorizing on a Map<Type,Count> with groupBy operation. We just need a new enum for that I think.

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/BulkWorkflow.java
efonsell: ok

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/BulkWorkflow.java
efonsell: Just reverted to how it was

nflow-engine/src/main/java/io/nflow/engine/internal/executor/WorkflowStateProcessor.java
```diff
@@ -204,9 +205,14 @@ private WorkflowInstance saveWorkflowInstanceState(StateExecutionImpl execution,
       logger.debug("No handler method defined for {}, clearing next activation", execution.getNextState());
       execution.setNextActivation(null);
     }
+    WorkflowState nextState = definition.getState(execution.getNextState());
+    if (definition.getSettings().wakeupParentWhenFinished && nextState.getType() == WorkflowStateType.end
+        && instance.parentWorkflowId != null) {
```
gmokki: It would be faster to reverse the order of checks here. Since looking at settings is slower than looking at instance/nextState varibles (but the difference is not measurable so I do not know why I wrote this comment)

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/BulkWorkflow.java
```diff
@@ -0,0 +1,149 @@
...
+  protected boolean splitWorkImpl(StateExecution execution, @SuppressWarnings("unused") JsonNode data) {
+    if (execution.getAllChildWorkflows().isEmpty()) {
+      throw new RuntimeException(
+          "No child workflows found - either add them before starting the parent or implement splitWorkflowImpl");
+    }
+    return true;
+  }
+
+  protected DateTime waitForChildrenUntil() {
+    return now().plusMinutes(1);
```
gmokki: I would like this to be much longer, preferably an 15min to 60min. When you have 1000 bulk workflows the constant load of just waking without child wakeup will just cause extra load (many wakeups per second). This should just be a workaround for lost wakeup signal. The 1 minute is usually too slow for actual progress without wakeups.

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/BulkWorkflow.java
```diff
@@ -0,0 +1,149 @@
...
+    if (completed == childWorkflows.size()) {
+      return moveToState(done, "All children completed");
+    }
+    long toStart = min(max(1, concurrency) - running, childWorkflows.size() - completed);
+    if (toStart > 0) {
+      childWorkflows.stream().filter(this::isInInitialState).limit(toStart).forEach(this::wakeup);
+      logger.info("Started " + toStart + " child workflows");
+    }
+    long progress = completed * 100 / childWorkflows.size();
+    return retryAfter(waitForChildrenToCompleteUntil(), "Waiting for child workflows to complete - " + progress + "% done");
```
gmokki: Did we have a limit on number of retries? This polling could be retried many times if there are many child workflows. Should we increase the limit or add some flag to make this retry not increase the counter.

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/DemoWorkflow.java
```diff
@@ -41,7 +42,7 @@ public String getDescription() {
   }
 
   public DemoWorkflow() {
-    super(DEMO_WORKFLOW_TYPE, State.begin, State.error);
+    super(DEMO_WORKFLOW_TYPE, State.begin, State.error, new WorkflowSettings.Builder().setWakeupParentWhenFinished(true).build());
```
gmokki: I believe the idea of the wakup parent was to avoid changing the child workflows. This just changes the code from the step to constructor, which is clearly better. But it would be good if parent could enable the flag on the child so that any workflow could be used as bulk child.

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/BulkWorkflow.java
efonsell: Increased to 1 hour

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/BulkWorkflow.java
efonsell: Changed retry limit for bulk workflow to Integer.MAX_VALUE. Is this ok or should we use moveToNextStateAfter instead?

nflow-engine/src/main/java/io/nflow/engine/internal/executor/WorkflowStateProcessor.java
efonsell: Fixed anyway

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/DemoWorkflow.java
efonsell: Remove wakeup flag from settings, added new state type wait, if parent is in wait state wake it up when child finishes

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
-## 5.3.4-SNAPSHOT (future release)
+## 5.4.0-SNAPSHOT (future release)
 
 **Highlights**
+- Introduce BulkWorkflow which can be used or extended to handle mass of child workflows without overloading the system
+- Allow child workflows to automatically wake up parent workflow instance when they enter end state (can be enabled via workflow settings)
```
gmokki: no longer "can be enable via workflow settings" but instead "when parent is in _wait_ state"

nflow-tests/src/test/java/io/nflow/tests/BulkWorkflowTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+@FixMethodOrder(NAME_ASCENDING)
+public class BulkWorkflowTest extends AbstractNflowTest {
+
+  @ClassRule
+  public static NflowServerRule server = new NflowServerRule.Builder().springContextClass(Configuration.class).build();
+
+  private static int workflowId;
+
+  private static int childrenCount;
```
gmokki: This shuold be a constant.

nflow-tests/src/test/java/io/nflow/tests/BulkWorkflowTest.java
```diff
@@ -0,0 +1,120 @@
...
+  static class Configuration {
+    // for component scanning only
+  }
+
+  @Test
+  public void t01_startDemoBulkWorkflow() {
+    CreateWorkflowInstanceRequest req = new CreateWorkflowInstanceRequest();
+    req.type = DEMO_BULK_WORKFLOW_TYPE;
+    req.stateVariables.put(BulkWorkflow.VAR_CONCURRENCY, 3);
+    List<Integer> childData = asList(1, 2, 3, 4, 5, 6, 7, 8, 9, 10);
```
gmokki: Replace `asList` with `IntStream.range(1, CHILD_COUNT).boxed().collect(toList())`

dahedgehog: If wanting to express the real count of the children with the CHILD_COUNT constant, rather use `IntStream.rangeClosed(1, CHILD_COUNT).boxed().collect(toList())`

nflow-tests/src/test/java/io/nflow/tests/BulkWorkflowTest.java
efonsell: Fixed

nflow-tests/src/test/java/io/nflow/tests/BulkWorkflowTest.java
efonsell: Fixed

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/WorkflowStateType.java
```diff
@@ -26,6 +26,12 @@
    */
   normal(false, WorkflowInstanceStatus.inProgress),
 
+  /**
+   * State for waiting something outside this instance to happen, including (but not limited to) child workflow instances to go to
+   * end state. When a child workflow finishes and parent is in wait state, parent is automatically woken up.
+   */
+  wait(false, WorkflowInstanceStatus.inProgress),
```
jsyrjala: Does nflow-explorer need changes to support this? e.g. own color for 'waiting' status?

efonsell: Own color added

PR #287, opened 2019-02-12
use select distinct when getting preserved action ids for history cleanup

efonsell:


nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -885,7 +885,7 @@ public int deleteWorkflowInstanceHistory(Integer workflowInstanceId, Integer his
       params.addValue("maxActionId", maxActionId);
       namedJdbc.update("delete from nflow_workflow_state where workflow_id = :workflowId and action_id <= :maxActionId", params);
       List<Integer> referredActionIds = namedJdbc.queryForList(
-          "select parent_action_id from nflow_workflow where parent_workflow_id = :workflowId", params, Integer.class);
+          "select distinct parent_action_id from nflow_workflow where parent_workflow_id = :workflowId", params, Integer.class);
```
jsyrjala: Update also CHANGELOG

PR #292, opened 2019-03-01
Show user defined content in ui

jsyrjala:
Users can now add custom Javascript functions to config.js. Those functions are invoked when rendering workflow definition and workflow instance page and output of the function is displayed on the page.

### DIFF
--- a/nflow-explorer/src/config.js
+++ b/nflow-explorer/src/config.js
@@ -36,4 +36,39 @@ var Config = function() {
     // max number of items to keep in memory
     maxHistorySize: 10000
   };
+
+  /**
+   * Generate custom content for workflow definition details page.
+   * Optional function that returns either a string or a Promise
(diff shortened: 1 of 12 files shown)


nflow-explorer/src/config.js
```diff
@@ -36,4 +36,39 @@ var Config = function() {
...
+   * - fetch additional data from external systems and display it in the in nFlow Explorer UI:
+   *   ```
+   *   return fetch('https://api.service.com/content/' + workflow.businessKey)
+   *            .then(result => {
+   *              return result.json()
+   *                .then(data => data.contentTitle)
+   *            })
+   *   ```
+   */
+  this.customInstanceContent = function(definition, workflow, parentWorkflow) {
```
efonsell:
```suggestion
  this.customInstanceContent = function(definition, workflow, parentWorkflow, childWorkflows) {
```

PR #296, opened 2019-03-06
Add support for adding ApplicationListener when starting netty

ttiurani:
Bonus: use plural form in varags.


jsyrjala:
Ok to merge if travis build passes.

PR #302, opened 2019-03-29
Fix WorkflowLifecycle start/stop/start

ttiurani:


ttiurani:
Instead of this, implemented a less-involved pause/resume functionality:

https://github.com/NitorCreations/nflow/pull/303

PR #309, opened 2019-05-14
Optimize instance loading etc

efonsell:


nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -513,6 +509,11 @@ public WorkflowInstance getWorkflowInstance(int id, Set<WorkflowInstanceInclude>
     return instance;
   }
 
+  public DateTime getWorkflowInstanceStartTime(int id) {
+    return toDateTime(
+        jdbc.queryForObject("select min(execution_start) from nflow_workflow_action where workflow_id = ?", Timestamp.class, id));
```
jsyrjala: Could we store this data to nflow_workflow table instead?

efonsell: We could, see https://github.com/NitorCreations/nflow/pull/310

Issue #311, opened 2019-05-15
Can not create Kotlin workflows


dahedgehog:
It's not possible to create pure Kotlin workflows extending WorkflowDefinition because WorkflowState must be and enum. This creates a clash between name property in Kotlin enum and name() method in WorkflowState interface. Example of non-functioning code:

```
enum class State(private val _type: WorkflowStateType, private val _desc: String) : WorkflowState {
    repeat(WorkflowStateType.start, "Start state"),
    error(WorkflowStateType.manual, "Error state");

    override fun getDescription() = _desc
    override fun getType() = _type
}
```

jsyrjala:
nFlow 5.6.0 was released with this feature.

https://github.com/NitorCreations/nflow/releases/tag/5.6.0

PR #312, opened 2019-05-15
Replace references to WorkflowDefinition with AbstractWorkflowDefinition

dahedgehog:
This will fix issue #311 by replacing references to WorkflowDefinition (which requires use of enum) with AbstractWorkflowDefinition (which works with any WorkflowState type classes).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
-## 5.5.1-SNAPSHOT (future release)
+## 5.6.0-SNAPSHOT (future release)
+
+**Highlights**
+- Support non-enum WorkflowStates to enable, for example, Kotlin workflow definitions by extending AbstractWorkflowDefinition.
 
 **Details**
 - Dependency and plugin updates:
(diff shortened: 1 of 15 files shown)


nflow-engine/src/main/java/io/nflow/engine/internal/workflow/WorkflowInstancePreProcessor.java
```diff
@@ -33,7 +33,7 @@ public WorkflowInstance process(WorkflowInstance instance) {
     }
     WorkflowInstance.Builder builder = new WorkflowInstance.Builder(instance);
```
jsyrjala: Could you update also CHANGELOG.md

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## 5.5.1-SNAPSHOT (future release)
```
jsyrjala: New feature is added => minor version bump here and in pom.xml files: (5.6.0-SNAPSHOT)

nflow-examples/spring-boot/full-stack-kotlin/src/workflow/ExampleWorkflow.kt
```diff
@@ -0,0 +1,45 @@
...
+@Component
+class ExampleWorkflow : KtWorkflowDefinition<State>(TYPE, repeat, error) {
+
+    init {
+        permit(repeat, repeat)
+    }
+
+    fun repeat(execution: StateExecution): NextAction {
+        println("Counter: " + execution.getVariable(VAR_COUNTER))
+        execution.setVariable(VAR_COUNTER, execution.getVar<Int>(VAR_COUNTER) + 1)
```
jsyrjala: Where is this `getVar<Int>` method? Is it some kind of kotlin magic thing?

jsyrjala: Found it.

dahedgehog: Yes, it's Kotlin extension function, and allows nice inline reified typing of the return value (as a type parameter to the function), instead of requiring to pass the type class as a parameter to the function. Kotlin magic...

ttiurani:
We want to put the two binary `.jar` files to version control? @dahedgehog

PR #316, opened 2019-05-25
Add NflowEngine class for starting nflow-engine easily

jsyrjala:
NflowEngine starts the neflow-engine and the caller doesn't need to use or worry about Spring setup.


jsyrjala:
PR contains unneeded stuff, will make a better one

PR #317, opened 2019-05-25
Add NflowEngine class for starting nflow-engine easily.

jsyrjala:
NflowEngine starts the nflow-engine and the caller doesn't need to use or worry about Spring setup.

### DIFF
--- a/nflow-engine/src/main/java/io/nflow/engine/NflowEngine.java
+++ b/nflow-engine/src/main/java/io/nflow/engine/NflowEngine.java
@@ -39,7 +39,8 @@ public class NflowEngine implements Runnable {
                        SQLVariants sqlVariants,
                        Collection<AbstractWorkflowDefinition<? extends WorkflowState>> workflowDefinitions) {
         AnnotationConfigApplicationContext ctx = new AnnotationConfigApplicationContext();
-        ctx.registerBean(DataSource.class, () -> dataSource);
+
+        ctx.registerBean("nflowDatasource", DataSource.class, () -> dataSource);
         ctx.registerBean(SQLVariants.class, () -> sqlVariants);
         ctx.register(EngineConfiguration.class, NflowEngineSpringConfig.class);
         ctx.refresh();
--- a/nflow-engine/src/main/java/io/nflow/engine/NflowEngine.java
+++ b/nflow-engine/src/main/java/io/nflow/engine/NflowEngine.java
@@ -24,9 +24,10 @@ import java.util.Collection;
  *
  * Use this if you want to start just the nflow-engine.
  */
-public class NflowEngine implements Runnable, AutoCloseable {
+public class NflowEngine implements AutoCloseable {
 
     private final WorkflowDispatcher workflowDispatcher;
+    private final AnnotationConfigApplicationContext ctx;
 
     public final ArchiveService archiveService;
     public final HealthCheckService healthCheckService;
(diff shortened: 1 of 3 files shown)


nflow-engine/src/main/java/io/nflow/engine/NflowEngine.java
```diff
@@ -0,0 +1,109 @@
...
+
+import javax.sql.DataSource;
+import java.util.Collection;
+
+/**
+ * NflowEngine starts up nflow-engine with given database and workflow definitions.
+ *
+ * Use this if you want to start just the nflow-engine.
+ */
+public class NflowEngine implements Runnable {
```
gmokki: I see very little benefit for implementing `Runnable` or is the startup really slow and should be parallelized. Anyway it is not wrong. Just thinking aloud.

gmokki: I would add `AutoClosable` interface so that if one uses the NflowEngine in some tests or other places the code to handle shutdown will be much cleaner.

nflow-engine/src/test/java/io/nflow/engine/NflowEngineTest.java
```diff
@@ -0,0 +1,121 @@
...
+     * This will enable scanning for io.nflow.engine.config.db which will create
+     * an extra data source, from e.g. H2DatabaseConfiguration.java
+     * The spring setup will fail because there are two competing dataSources.
+     *
+     * This uses system property `spring.profiles.active`
+     * to override env variable `SPRING_PROFILES_ACTIVE`.
+     */
+    @BeforeEach
+    public void setup() {
+        System.setProperty("spring.profiles.active", "nflow.db.dummy");
```
gmokki: What if you would register the wanted DataSource as Primary?
Maybe using `https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/config/BeanDefinitionCustomizer.html` as extra argument in the `registerBean`

nflow-engine/src/main/java/io/nflow/engine/NflowEngine.java
jsyrjala: Typically you would run the NflowEngine in a thread and you can pass this class directly to thread.

jsyrjala: Implemented

jsyrjala: Actually WorkflowLifecycle already started the dispatcher in a thread. Removed Runnable from NflowEngine.java. Also replaced NflowEngine#shutdown() with NflowEngine#close().

nflow-engine/src/test/java/io/nflow/engine/NflowEngineTest.java
jsyrjala: If I do that the end result seems to be that Spring finds 0 suitable dataSource beans and fails.

nflow-engine/src/main/java/io/nflow/engine/NflowEngine.java
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * Pauses a running nFlow engine.
+     */
+    public void pause() {
+        workflowLifecycle.pause();
+    }
+
+    /**
+     * Resumes paused the nFlow engine.
```
efonsell:
```suggestion
     * Resumes a paused nFlow engine.
```

nflow-engine/src/main/java/io/nflow/engine/NflowEngine.java
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * Returns true if the nFlow engine is currently running.
+     */
+    public boolean isRunning() {
+        return workflowLifecycle.isRunning();
+    }
+
+    /**
+     * Shuts down the nFlow engine.
```
efonsell: How? Gracefully, and returns when done? Or just triggers the shutdown and returns immediately?

jsyrjala: Updated javadocs

PR #321, opened 2019-05-28
No mariadb for mysql

gmokki:
This is the only way to have both maridb and mysql jdbc drivers in the classpath at the same time


efonsell:
add changelog entry

PR #324, opened 2019-06-02
separate definition scanning from definition service

efonsell:


nflow-engine/src/main/java/io/nflow/engine/service/WorkflowDefinitionService.java
```diff
@@ -33,31 +25,17 @@
...
-  public void setWorkflowDefinitions(Collection<AbstractWorkflowDefinition<? extends WorkflowState>> workflowDefinitions) {
-    for (AbstractWorkflowDefinition<? extends WorkflowState> wd : workflowDefinitions) {
-      addWorkflowDefinition(wd);
-    }
-  }
-
-  @Autowired(required = false)
-  public void setWorkflowDefinitions(@NFlow AbstractResource nflowNonSpringWorkflowsListing) {
-    this.nonSpringWorkflowsListing = nflowNonSpringWorkflowsListing;
+    this.autoInit = env.getRequiredProperty("nflow.autoinit", Boolean.class);
```
gmokki: Could we combine the boolean logic of `autoInit=false` => `persistWorkflowDefinitions=false` already in constructor and not store the autoInit as field?

nflow-engine/src/main/java/io/nflow/engine/service/WorkflowDefinitionService.java
```diff
@@ -78,42 +56,28 @@ public void setWorkflowDefinitions(@NFlow AbstractResource nflowNonSpringWorkflo
...
-   * @throws ReflectiveOperationException when the workflow definition can not be instantiated.
+   * Persist all loaded workflow definitions if needed.
    */
-  public void postProcessWorkflowDefinitions() throws IOException, ReflectiveOperationException {
-    if (nonSpringWorkflowsListing == null) {
-      logger.info("No non-Spring workflow definitions");
-    } else {
-      initNonSpringWorkflowDefinitions();
-    }
+  public void postProcessWorkflowDefinitions() {
```
gmokki: Now that we store definitions in the addWorkflowDefinition already, we should just remove this `postProcessWorkflowDefinitions` method.
It will also nicely clean up some dependencies from other components.

nflow-engine/src/main/java/io/nflow/engine/service/WorkflowDefinitionService.java
efonsell: No, since we need to persist in different method depending on autoInit.

nflow-engine/src/main/java/io/nflow/engine/service/WorkflowDefinitionService.java
efonsell: We can't remove it since we need to persist in different phase depending on autoinit, but I moved persisting logic completely to workflow definition service.

nflow-engine/src/main/java/io/nflow/engine/service/WorkflowDefinitionService.java
```diff
@@ -71,49 +52,45 @@ public void setWorkflowDefinitions(@NFlow AbstractResource nflowNonSpringWorkflo
 
   /**
    * Return all managed workflow definitions.
+   *
    * @return List of workflow definitions.
    */
   public List<AbstractWorkflowDefinition<? extends WorkflowState>> getWorkflowDefinitions() {
-    return new ArrayList<>(workflowDefinitions.values());
+    synchronized (workflowDefinitions) {
```
jsyrjala: Is synchronized needed here? workflowDefinitions is already a synchronizedMap.

nflow-engine/src/test/java/io/nflow/engine/service/WorkflowDefinitionServiceWithSpringTest.java
```diff
@@ -50,20 +50,22 @@
 @DirtiesContext
 public class WorkflowDefinitionServiceWithSpringTest {
 
+  static WorkflowDefinitionService workflowDefinitionService;
```
jsyrjala: You could try injecting (or @Autowired) the bean here, like WorkflowDefinitionSpringBeanScanner is done at line 160?

```
@Inject
private WorkflowDefinitionService workflowDefinitionService;
```

nflow-engine/src/main/java/io/nflow/engine/service/WorkflowDefinitionService.java
efonsell: Yes it is, "It is imperative that the user manually synchronize on the returned map when iterating over any of its collection views"

nflow-engine/src/test/java/io/nflow/engine/service/WorkflowDefinitionServiceWithSpringTest.java
efonsell: Done

PR #325, opened 2019-06-03
Switch travis from ubuntu 14.04 to 16.04

gmokki:


jsyrjala:
Travis build fails now with JDK8.

PR #338, opened 2019-08-24
Replace FindBugs with SpotBugs, fix some warnings, refactor

efonsell:
WIP


.travis.yml
```diff
@@ -2,7 +2,6 @@ dist: trusty
 language: java
 jdk:
   - openjdk11
-  - oraclejdk8
```
gmokki: Are we sure we want to drop java8 support? The adoptopenjdk will keep supporting it at least until end of 2022

nflow-engine/src/main/java/io/nflow/engine/internal/dao/ExecutorDao.java
```diff
@@ -67,13 +67,6 @@ public ExecutorDao(SQLVariants sqlVariants, @NFlow JdbcTemplate nflowJdbcTemplat
     this.hostMaxLength = env.getProperty("nflow.executor.host.length", Integer.class, -1);
   }
 
-  private int getHostMaxLength() {
```
gmokki: Should we instead make this a global helper that takes a table name and column name?

nflow-engine/src/main/java/io/nflow/engine/internal/dao/TableMetadataChecker.java
```diff
@@ -68,7 +70,7 @@ public void ensureCopyingPossible(String sourceTable, String destinationTable) {
     @Override
     public Map<String, ColumnMetadata> extractData(ResultSet rs) throws SQLException {
       ResultSetMetaData metadata = rs.getMetaData();
-      Map<String, ColumnMetadata> metadataMap = new LinkedHashMap<>();
+      Map<String, ColumnMetadata> metadataMap = new LinkedHashMap<>(metadata.getColumnCount(), 1.0f);
```
gmokki: Why add an explicit fill factor? And the closer to `1` the value is the more likely it is that the map will become linked list. We're not trying to save any memory here so I think the deafult `0.75` is good here.

nflow-engine/src/main/java/io/nflow/engine/internal/workflow/StateExecutionImpl.java
```diff
@@ -103,18 +104,16 @@ public String getVariable(String name) {
   @Override
   @SuppressWarnings("unchecked")
   public <T> T getVariable(String name, Class<T> type, T defaultValue) {
-    if (instance.stateVariables.containsKey(name)) {
-      return (T) objectMapper.convertToObject(type, name, instance.stateVariables.get(name));
+    String value = instance.stateVariables.get(name);
```
gmokki: Does this change the semantincs?
Previously an explicit null value in db would not return `defaultValue` but `null` form this method.
I'm not against the semantics change, but maybe we should at least document it in changes.

nflow-engine/src/main/java/io/nflow/engine/service/WorkflowDefinitionService.java
```diff
@@ -54,7 +56,9 @@ public WorkflowDefinitionService(WorkflowDefinitionDao workflowDefinitionDao, En
    * @return List of workflow definitions.
    */
   public List<AbstractWorkflowDefinition<? extends WorkflowState>> getWorkflowDefinitions() {
-    return new ArrayList<>(workflowDefinitions.values());
+    synchronized (workflowDefinitions) {
```
gmokki: Why pessimize the get operations?
- If this method is not called often then it does not matter and the new code might be a bit simpler to read.
- If this method is called often then we should precalculate the `workflowDefinitions.values()` in an unmodifiableList and return that

nflow-engine/src/test/java/io/nflow/engine/internal/executor/WorkflowStateProcessorTest.java
```diff
@@ -100,10 +99,9 @@
 public class WorkflowStateProcessorTest extends BaseNflowTest {
 
   /**
-   * TODO junit5 doesn't have global timeouts yet
-   * https://stackoverflow.com/questions/47041313/junit-5-global-timeout
-   * @Rule
-   * public Timeout timeoutPerMethod = Timeout.seconds(5);
+   * TODO junit5 doesn't have global timeouts yet https://stackoverflow.com/questions/47041313/junit-5-global-timeout
```
gmokki: Is this comment still needed? I think I added timeouts back to many tests already

pom.xml
```diff
@@ -157,7 +153,7 @@
     <maven-surefire.version>2.22.2</maven-surefire.version>
     <metrics.version>4.1.0</metrics.version>
     <mockito.version>3.0.0</mockito.version>
-    <mssql.version>7.4.1.jre8</mssql.version>
+    <mssql.version>7.4.1.jre11</mssql.version>
```
gmokki: If we want to keep the profiles we could override this for if java11 is detected?

nflow-engine/src/main/java/io/nflow/engine/internal/dao/TableMetadataChecker.java
efonsell: Updated code to use double initial size and default load factor (double size to avoid rehashing)

nflow-engine/src/main/java/io/nflow/engine/internal/workflow/StateExecutionImpl.java
efonsell: If stateVariable value is null, objectMapper.convertToObject throws NPE, so it does not change anything that was working before. Instead it fixes the case when null values has been stored for a state variable, and uses the default value instead as I would expect.

nflow-engine/src/main/java/io/nflow/engine/service/WorkflowDefinitionService.java
efonsell: Optiomized for get

pom.xml
```diff
@@ -172,6 +168,10 @@
     <!-- 0.9.11 not used because https://trello.com/c/ZEcAoiSq/406-remove-05-mb-of-exceptions-farted-out-when-starting-nflow-jetty -->
     <reflections.version>0.9.11</reflections.version>
     <slf4j.version>1.7.29</slf4j.version>
+    <spotbugs.version>3.1.12</spotbugs.version>
+    <spotbugs-annotations.version>3.1.12</spotbugs-annotations.version>
+    <spotbugs-contrib.version>7.4.6</spotbugs-contrib.version>
+    <spotbugs-maven-plugin.version>3.1.12.1</spotbugs-maven-plugin.version>
```
jsyrjala: `3.1.12.2` is the latest version

PR #341, opened 2019-09-24
Version bumps

gmokki:


pom.xml
```diff
@@ -161,19 +161,19 @@
...
+    <node.version>v10.16.3</node.version> <!-- https://nodejs.org/en/download/releases/ -->
+    <npm.version>6.11.3</npm.version> <!-- https://github.com/npm/cli/releases -->
     <nitor-matchers.version>1.3</nitor-matchers.version>
-    <postgresql.version>42.2.6</postgresql.version>
+    <postgresql.version>42.2.8</postgresql.version>
     <postgresql-embedded.version>2.10</postgresql-embedded.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <reactor.netty.version>0.8.9.RELEASE</reactor.netty.version>
+    <reactor.netty.version>0.8.11.RELEASE</reactor.netty.version>
```
ttiurani: We are depending on:

https://mvnrepository.com/artifact/org.springframework/spring-webflux/5.1.9.RELEASE

which has `reactor.netty` `0.8.10.RELEASE`. Not sure if having a mismatch there is a good idea? @gmokki

gmokki: https://github.com/reactor/reactor-netty/releases/tag/v0.8.11.RELEASE looks like a pretty safe release to me, they even recommend it to all 0.8.x users

ttiurani: Ah, true, even the spring-boot people brought in 0.8.11 with 5.1.9:

https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-reactor-netty/2.1.8.RELEASE

So nevermind.

PR #343, opened 2019-10-10
add index for polling

efonsell:


nflow-engine/src/main/resources/scripts/db/sqlserver.create.ddl.sql
```diff
@@ -48,12 +48,18 @@ create table nflow_workflow_action (
   constraint nflow_workflow_action_uniq unique (workflow_id, id)
 );
 
+create index nflow_workflow_action_workflow on nflow_workflow_action(workflow_id);
```
jsyrjala: There should be also a update.sql that also adds these things.

jsyrjala: And there is such a update.sql

jsyrjala:
CHANGELOG  entry is missing.

PR #347, opened 2019-11-10
Improvements and fixes for h2 and mssql ddl scripts

tomkiljo:
Improvements for h2 create ddl, added trigger "if not exists" clause.
Improvements for mssql create ddl, added mssql style "if not exists" clauses.
Fix h2 and mssql latest update ddl (update-5.6.0-X) column `started` populating.

MSSQL scripts verified with MSSQL 2017-CU14-ubuntu and AzureSQL.


gmokki:
Thanks for fixing the upgrade scrips. Looks good.
This shows clearly that we do not currently test the upgrades, although rarely there are any bigger changes like this that need more complex logic.

I'm not sure that if we should have the if not modified -logic, since no one should run the create scripts multiple times. But at least this PR matches what is done in other places so removing them would clearly be a separate PR if we want to do it.

PR #349, opened 2019-11-14
Add nflow.db.disable_batch_updates, default to false

ttiurani:
This makes it possible to revert to the multi update model even if the database needs it. This needs to be done to prevent MariaDB/Galera from having double executions.

### DIFF
--- a/nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
+++ b/nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
@@ -98,6 +98,7 @@ public class WorkflowInstanceDao {
   private final long workflowInstanceQueryMaxResultsDefault;
   private final long workflowInstanceQueryMaxActions;
   private final long workflowInstanceQueryMaxActionsDefault;
+  private final boolean disableBatchUpdates;
   int instanceStateTextLength;
   int actionStateTextLength;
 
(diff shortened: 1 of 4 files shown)


nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -128,6 +129,10 @@ public WorkflowInstanceDao(SQLVariants sqlVariants,
     // In one deployment, FirstColumnLengthExtractor returned 0 column length (H2), so allow explicit length setting.
     instanceStateTextLength = env.getProperty("nflow.workflow.instance.state.text.length", Integer.class, -1);
     actionStateTextLength = env.getProperty("nflow.workflow.action.state.text.length", Integer.class, -1);
+    disableBatchUpdates = env.getProperty("nflow.db.disable_batch_updates", Boolean.class);
```
efonsell:
```suggestion
    disableBatchUpdates = env.getRequiredProperty("nflow.db.disable_batch_updates", Boolean.class);
```

efonsell:
Release notes missing, also remember to document new configuration option to Wiki

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
ttiurani: Did this fix, but force pushed

PR #351, opened 2019-11-18
Drop index that is only used for archiving purposes. 

gmokki:
It slows down every nflow event


efonsell:
Remove this from sqlserver script and drop it in upgrade script:
```
if not exists (select 1 from sys.indexes where name='nflow_archive_workflow_activation')
create index nflow_archive_workflow_activation on nflow_archive_workflow(next_activation, modified);
```

PR #352, opened 2019-11-22
Refactor workflow and workflow action id from int to long

gmokki:


nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -1,8 +1,6 @@
 package io.nflow.engine.internal.dao;
 
-import static io.nflow.engine.internal.dao.DaoUtil.firstColumnLengthExtractor;
-import static io.nflow.engine.internal.dao.DaoUtil.getInt;
-import static io.nflow.engine.internal.dao.DaoUtil.toTimestamp;
+import static io.nflow.engine.internal.dao.DaoUtil.*;
```
efonsell: expand

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -85,7 +83,7 @@
 public class WorkflowInstanceDao {
 
   private static final Logger logger = getLogger(WorkflowInstanceDao.class);
-  static final Map<Integer, Map<String, String>> EMPTY_ACTION_STATE_MAP = Collections.<Integer, Map<String, String>> emptyMap();
+  static final Map<Long, Map<String, String>> EMPTY_ACTION_STATE_MAP = Collections.emptyMap();
```
efonsell: static import

nflow-engine/src/test/java/io/nflow/engine/internal/executor/WorkflowDispatcherTest.java
```diff
@@ -303,7 +303,7 @@ public void dispatcherLogsWarningWhenAllThreadsArePotentiallyStuck() throws Thro
     class DispatcherLogsWarning extends MultithreadedTestCase {
       public void threadDispatcher() throws InterruptedException {
         when(workflowInstances.pollNextWorkflowInstanceIds(anyInt()))
-            .thenAnswer(waitForTickAndAnswer(2, Collections.<Integer> emptyList(), this));
+            .thenAnswer(waitForTickAndAnswer(2, Collections.emptyList(), this));
```
efonsell: static import

efonsell:
Added some minor comments. In general, if we would use long constants instead of magic numbers in tests we would not need to put those Ls everywhere...

PR #354, opened 2019-11-23
Add priority to workflow instance

ttiurani:
Adds `priority` two byte integer to the `nflow_workflow` table. When an executor chooses from many available scheduled workflow instances it primarily (unfairly) schedules the workflow instance with the larger priority value, and for workflows with the same priority, the one scheduled first. Priority defaults to 0 and can also be negative. (The difference in priority integer value is not significant: 1 vs 2 is treated identically to  -1000 vs 1000 – in both cases latter is executed before the former.)

Database migration is documented with options to either do it with one alter table (for reasonably small workflow instance count) or in phases.

https://github.com/NitorCreations/nflow/pull/355 adds support for specifying a default priority in the workflow definition, which is the most typical use case.


nflow-engine/src/main/java/io/nflow/engine/workflow/instance/WorkflowInstance.java
```diff
@@ -75,6 +75,11 @@
    */
   public final String type;
 
+  /**
+   * The priority of the workflow instance. Defaults to 0 and can also be negative.
```
jsyrjala: Could you document which direction, higher or lower priority numbers gets more attention and processing power from the nflow?

nflow-engine/src/main/java/io/nflow/engine/workflow/instance/WorkflowInstance.java
```diff
@@ -366,6 +374,16 @@ public Builder setType(String type) {
       return this;
     }
 
+    /**
+     * Set the priority of the workflow instance.
+     * @param priority The priority.
```
jsyrjala: Document that higher number gives higher priority?

nflow-rest-api-common/src/main/java/io/nflow/rest/v1/msg/ListWorkflowInstanceResponse.java
```diff
@@ -27,6 +27,9 @@
   @ApiModelProperty(value = "Workflow definition type", required = true)
   public String type;
 
+  @ApiModelProperty(value = "Workflow instance priority", required = true)
```
jsyrjala: Documenting what it means to have higher number here?

nflow-engine/src/main/resources/scripts/db/db2.create.ddl.sql
```diff
@@ -4,6 +4,7 @@ create table nflow_workflow (
   id int primary key generated always as identity,
```
jsyrjala: Changes are missing from update sql files.

nflow-engine/src/main/resources/scripts/db/db2.create.ddl.sql
```diff
@@ -4,6 +4,7 @@ create table nflow_workflow (
...
   status varchar(32) not null check (status in ('created', 'executing', 'inProgress', 'finished', 'manual')),
   type varchar(64) not null,
+  priority int not null default 0,
```
jsyrjala: There is no index changes related to priority field?

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -186,13 +187,17 @@ private int insertWorkflowInstanceWithCte(WorkflowInstance instance) {
     }
   }
 
+  private int getInstancePriority(WorkflowInstance instance) {
+    return instance.priority != null ? instance.priority.intValue() : 0;
```
efonsell: if the column is not null in database, why we need this null check?

efonsell:
I think it would be nice to have the default priority for new instances defined in the workflow definition (which would have default value 0)

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
ttiurani: The method is used for inserting a new workflow, and if `setPriority` has not been called in the builder, the value is null. I thought this is the best place to set this, because there will (in a later PR) come updating a wf priority, where null value in the instance is useful to determine that no changes to priority should be made.

nflow-engine/src/main/resources/scripts/db/db2.create.ddl.sql
ttiurani: No, because ORDER BY index will only matter if the executor has to choose from thousands of instances to be executed right now, which is not a case that warrants increasing the size of the index.

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
efonsell: Ok

nflow-engine/src/main/resources/scripts/db/db2.create.ddl.sql
ttiurani: Added those

CHANGELOG.md
```diff
@@ -14,6 +15,7 @@
 
 **Details**
 - `nflow-engine`
+  - Add `priority` two byte integer to the `nflow_workflow` table. When an executor chooses from many available scheduled workflow instances it primarily (unfairly) schedules the workflow instance with the larger priority value, and for workflows with the same priority, the one scheduled first. Priority defaults to 0 and can also be negative. Requires database migration.
```
efonsell: Maybe mention here that priority can be set on definition (default) and instance (override) level?

efonsell: Or should I update this in my pr...

ttiurani: yeah, thought that this would be a two PR thing, as you already did your own :)

PR #355, opened 2019-11-24
Set default priority for new workflow instances per workflow definition

efonsell:


nflow-rest-api-common/src/main/java/io/nflow/rest/v1/msg/ListWorkflowDefinitionResponse.java
```diff
@@ -41,6 +41,9 @@
     @ApiModelProperty(value = "Delay after which workflow instance history (actions, states) can be deleted from database", required = false)
     public Integer historyDeletableAfterHours;
 
+    @ApiModelProperty(value = "Default priority for new workflow instances", required = false)
```
ttiurani: Hmm, this is required=false but the raw type short?

efonsell: Fixed

PR #356, opened 2019-11-26
Version bump

gmokki:
I bumped some test dependencies (junit/spotbugs) to pre-releaseses. They cannot have big impacts on the produced artifacts, but might better reveal/test bugs from the current code.


efonsell:
Why beta and rc versions?

gmokki:
I tried to reason why in the PR description. Basically only test code is rc/beta. And it is in the hopes that it helps better spot new bugs or help write junit tests easier.

PR #359, opened 2019-11-28
Update deps in examples

efonsell:


efonsell:
Need to fix examples so that workflow instance is inserted in startup, otherwise example does not work as described here:
https://github.com/NitorCreations/nflow/wiki/Spring-Boot-guide
The old way of inserting instance in @PostConstruct does not work with 6.0.0, the definitions are not scanned by that time so the insert fails.

efonsell:
Inserting instances with ApplicationRunner seems to work

PR #360, opened 2019-11-29
Check state variable value length when setting variable.

efonsell:
Abbreviate or throw exception based on config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 **Details**
 - `nflow-engine`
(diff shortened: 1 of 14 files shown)


gmokki:
Abbreviate is a good idea. But if it is not enabled, should we schedule the workflow to occur later?
If we just throw exception then it will just usually cause immediate retry of the step

efonsell:
> Abbreviate is a good idea. But if it is not enabled, should we schedule the workflow to occur later?
> If we just throw exception then it will just usually cause immediate retry of the step

It will use the normal failure delay of the step, and you can implement custom handling for the exception in the step if you like.

PR #366, opened 2020-01-09
Minor fixes

gmokki:


nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -673,7 +673,7 @@ public int compareTo(OptimisticLockKey other) {
     }
     conditions.add("executor_group = :executor_group");
     params.addValue("executor_group", executorInfo.getExecutorGroup());
-    sql += " where " + collectionToDelimitedString(conditions, " and ") + " order by created desc";
+    sql += " where " + collectionToDelimitedString(conditions, " and ") + " order by id desc";
```
efonsell: Or maybe keep existing behaviour in most cases and fix the problematic special case only:
```suggestion
    sql += " where " + collectionToDelimitedString(conditions, " and ") + " order by created desc, id desc";
```

gmokki: I would think that the autogenerated id + created timestamp should correlate exactly?
Both are increasing numbers based on the insertion time

efonsell: True, I was thinking about a case where you could set the created timestamp to some other value. But I guess we do not allow that.

PR #371, opened 2020-01-21
Tune workflow polling.

gmokki:
- Fetch optimistic locked list of workflow instance ids in separate transaction from the modify phase. This should reduce the risk of deadlocks in databases.
- Also rework code so that if batch returns unsupported values we switch to non-batch usage.

### DIFF
--- a/nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
+++ b/nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
@@ -61,7 +61,6 @@ import org.springframework.jdbc.support.KeyHolder;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.TransactionStatus;
 import org.springframework.transaction.annotation.Transactional;
-import org.springframework.transaction.support.TransactionCallback;
 import org.springframework.transaction.support.TransactionCallbackWithoutResult;
 import org.springframework.transaction.support.TransactionTemplate;
 import org.springframework.util.Assert;
@@ -575,8 +574,8 @@ public class WorkflowInstanceDao {
       return null;
     });
     if (ids.isEmpty()) {
-      logger.debug("Race condition in polling workflow instances detected. "
-              + "Multiple pollers using same name ({})", executorInfo.getExecutorGroup());
+      throw new PollingRaceConditionException("Race condition in polling workflow instances detected. "
+              + "Multiple pollers using same name (" + executorInfo.getExecutorGroup() + ")");
     }
     return ids;
   }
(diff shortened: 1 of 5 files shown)


nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -553,69 +552,63 @@ String whereConditionForInstanceUpdate() {
...
+     for (OptimisticLockKey instance : instances) {
+       batchArgs.add(new Object[] { instance.id, sqlVariants.tuneTimestampForDb(instance.modified) });
+       ids.add(instance.id);
+     }
+     int[] updateStatuses = jdbc
+             .batchUpdate(updateInstanceForExecutionQuery() + " where id = ? and modified = ? and executor_id is null", batchArgs);
+     Iterator<Long> idIt = ids.iterator();
+     for (int status : updateStatuses) {
+       idIt.next();
+       if (status == 1) {
```
efonsell: I guess it should never happen, but what if status > 1? Now it's handled like status == 1, except that it makes to more checks for the value. So this could be status > 0 here?

nflow-engine/src/test/java/io/nflow/engine/internal/dao/WorkflowInstanceDaoTest.java
```diff
@@ -700,25 +700,6 @@ private WorkflowInstanceDao preparePostgreSQLDao(JdbcTemplate jdbcTemplate) {
     return d;
   }
 
-  @Test
```
efonsell: Why is this removed? Is the race condition not happening anymore?

nflow-engine/src/test/java/io/nflow/engine/internal/dao/WorkflowInstanceDaoTest.java
```diff
@@ -1018,7 +999,7 @@ public Poller(WorkflowInstanceDao dao, int batchSize) {
     public void run() {
       try {
         returnSize = min(returnSize, dao.pollNextWorkflowInstanceIds(batchSize).size());
-      } catch (PollingRaceConditionException ex) {
+      } catch (PollingBatchException ex) {
```
efonsell: Is this change still valid?

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
gmokki: Good point. Rewrote the logic a bit to be shorter and consider values > 1 a failure

PR #374, opened 2020-01-26
Archiving rewrite and support for deleting old data from main and archive tables

gmokki:
work in progress


nflow-engine/src/main/java/io/nflow/engine/internal/dao/ArchiveDao.java
```diff
@@ -64,34 +74,42 @@ public int archiveWorkflows(Collection<Long> workflowIds) {
...
-            "select " + columns + " from nflow_workflow_state where workflow_id in " + workflowIdParams);
+            "select " + columns + " from nflow_workflow_state where workflow_id in " + workflowIdParams + sqlVariants.forUpdateInnerSelect());
   }
 
-  private void deleteWorkflows(String workflowIdParams) {
-    jdbc.update("delete from nflow_workflow_state where workflow_id in " + workflowIdParams);
-    jdbc.update("update nflow_workflow set root_workflow_id=null, parent_workflow_id=null, parent_action_id=null " +
+  private int deleteWorkflows(TablePrefix table, String workflowIdParams) {
+    jdbc.update("update " + table.nameOf("workflow") + " set root_workflow_id=null, parent_workflow_id=null, parent_action_id=null " +
             "where id in " + workflowIdParams + " and (root_workflow_id is not null or parent_workflow_id is not null)");
```
efonsell: check if parent_action_id is null as well?

nflow-engine/src/main/resources/scripts/db/sqlserver.create.ddl.sql
```diff
@@ -54,17 +54,6 @@ create table nflow_workflow_action (
...
-  foreign key (parent_workflow_id, parent_action_id) references nflow_workflow_action (workflow_id, id) on update no action;
-
-if not exists (select 1 from sys.indexes where name='nflow_workflow_parent')
-create index nflow_workflow_parent on nflow_workflow(parent_workflow_id, parent_action_id) where parent_workflow_id is not null;
-
-if not exists (select 1 from sys.foreign_keys where name='fk_workflow_root')
-alter table nflow_workflow add constraint fk_workflow_root
-  foreign key (root_workflow_id) references nflow_workflow (id) on update no action;
-
 if not exists (select 1 from sys.indexes where name='nflow_workflow_root')
```
efonsell: remove this as well?

nflow-engine/src/main/java/io/nflow/engine/service/ArchiveService.java
```diff
@@ -40,34 +50,162 @@ public ArchiveService(ArchiveDao archiveDao) {
...
    */
-  @SuppressFBWarnings(value = "BAS_BLOATED_ASSIGNMENT_SCOPE", justification = "periodicLogger is defined in correct scope")
+  @Deprecated
   public int archiveWorkflows(DateTime olderThan, int batchSize) {
-    Assert.notNull(olderThan, "olderThan must not be null");
-    Assert.isTrue(batchSize > 0, "batchSize must be greater than 0");
-    archiveDao.ensureValidArchiveTablesExist();
-    log.info("Archiving starting. Archiving passive workflows older than {}, in batches of {}.", olderThan, batchSize);
+    return cleanupWorkflows(new Builder()
+            .setArchiveWorkflowsOlderThan(new Duration(olderThan, Instant.now()))
```
efonsell: Converting DateTime to duration and then back to DateTime later for the query might change the actual value a bit - I would keep it as DateTime to give the caller full control about the limit.

gmokki: This is only a backward compatible api. My plan is to delete this method in the next major release.
I would think that the new way of specifying how old, in relative time, instances will be operated on will be an easier to configure/use api.

efonsell: Ok

nflow-engine/src/main/java/io/nflow/engine/service/ArchiveService.java
```diff
@@ -40,34 +50,162 @@ public ArchiveService(ArchiveDao archiveDao) {
...
     } while (true);
 
-    log.info("Archiving finished. Archived {} workflows in {} seconds.", archivedWorkflowsTotal, stopWatch.getTime() / 1000);
-    return archivedWorkflowsTotal;
+    log.info("{} finished. Operated on {} workflows in {} seconds.", type, totalWorkflows, stopWatch.getTime() / 1000);
+    return totalWorkflows;
+  }
+
+  public static class ArchiveConfiguration {
+    private final Duration deleteArchivedWorkflowsOlderThan;
```
efonsell: these fields could be public? at least package protected to avoid synthetic accessors

PR #377, opened 2020-02-17
add workflow type filter to maintenance config

efonsell:


nflow-engine/src/main/java/io/nflow/engine/internal/dao/MaintenanceDao.java
```diff
@@ -65,12 +66,20 @@ private String getStateColumns() {
...
+        + sqlVariants.dateLtEqDiff("modified", "?") + getWorkflowTypeCondition(workflowTypes) + " order by id asc",
+        maxWorkflows);
     return jdbc.queryForList(sql, Long.class, sqlVariants.toTimestampObject(before));
   }
 
+  private String getWorkflowTypeCondition(Set<String> workflowTypes) {
+    if (workflowTypes.isEmpty()) {
+      return "";
+    }
+    return " and type in (" + join(workflowTypes, ",") + ")";
```
gmokki: This is allows sql injection. Random text from outside. At least the rest interface should filter out types that are invalid (or just characters that are not in the basic set)

nflow-rest-api-common/src/main/java/io/nflow/rest/v1/converter/MaintenanceConverter.java
```diff
@@ -27,7 +31,11 @@ public MaintenanceConfiguration convert(MaintenanceRequest request) {
   }
 
   private ConfigurationItem createConfig(MaintenanceRequestItem config) {
-    return new ConfigurationItem.Builder().setOlderThanPeriod(config.olderThanPeriod).setBatchSize(config.batchSize).build();
+    return new ConfigurationItem.Builder() //
+        .setOlderThanPeriod(config.olderThanPeriod) //
+        .setBatchSize(config.batchSize) //
+        .setWorkflowTypes(ofNullable(config.workflowTypes).orElseGet(Collections::emptySet)) //
```
gmokki: `.orElse(emptySet())` is faster. No need for lambda.

nflow-engine/src/main/java/io/nflow/engine/internal/dao/MaintenanceDao.java
gmokki: Similar problem seems to be in wakeup rest call that eventually calls `WorkflowInstanceDao.addExpectedStatesToQueryAndUpdate`

nflow-rest-api-common/src/main/java/io/nflow/rest/v1/converter/MaintenanceConverter.java
efonsell: I though config.workTypes would usually be present so calling emptySet() could be avoided when it is not really needed.

gmokki: Yes. That is the usual logic. If you would do a `new ArrayList()` then it would be better to use `orElseGet`. But since `emptySet()` is a constant variable there is no harm in passing the constant value. For example `.orElse(-1)` vs `.orElseGet(() -> -5)`

nflow-engine/src/main/java/io/nflow/engine/internal/dao/MaintenanceDao.java
efonsell: Added config validation, now only supported workflow definition types are allowed

nflow-engine/src/main/java/io/nflow/engine/service/MaintenanceService.java
```diff
@@ -76,6 +83,17 @@ public MaintenanceResults cleanupWorkflows(MaintenanceConfiguration configuratio
     return builder.build();
   }
 
+  private void validateConfiguration(MaintenanceConfiguration configuration) {
+    Stream.of(configuration.archiveWorkflows, configuration.deleteArchivedWorkflows, configuration.deleteWorkflows) //
+        .filter(Objects::nonNull) //
+        .flatMap(configItem -> configItem.workflowTypes.stream()) //
+        .forEach(workflowType -> {
```
gmokki: You could do more stream/optional stuff with the following, but I'm not sure if it looks nicer or not
```java
            .filter(workflowType -> workflowDefinitionService.getWorkflowDefinition(workflowType) == null)
            .findAny()
            .ifPresent(workflowType -> { throw new IllegalArgumentException("Workflow type " + workflowType + " is not valid"); });
```

efonsell: Yes it looks nicer.

PR #380, opened 2020-02-18
Built-in MaintenanceWorkflow and abstract CronWorkflow to be extended by other long-running workflows

gmokki:


nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public String getDescription() {
+      return name();
+    }
+  }
+
+  protected CronWorkflow(String type, WorkflowSettings settings) {
+    super(type, schedule, handleFailure, settings);
+    permit(schedule, doWork);
+    permit(doWork, schedule, handleFailure);
```
efonsell: handleFailure not needed here, it is already set as defaul error state above

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+    permit(handleFailure, schedule);
```
efonsell: the error state for handleFailure should be failed though?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  protected CronWorkflow(String type) {
+    this(type, new Builder()
```
efonsell: no need to break lines here?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+            .setHistoryDeletableAfter(weeks(1))
+            .build());
+  }
+
+  public NextAction schedule(@SuppressWarnings("unused") StateExecution execution, @StateVar(value = VAR_SCHEDULE, readOnly = true) String cron) {
```
efonsell: execution is not unused?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+  protected DateTime nextActivationTime(String cron, DateTime activationTime) {
+    DateTime next = new DateTime(new CronSequenceGenerator(cron).next(activationTime.toDate()));
+    return next.isBeforeNow() ? now() : next;
+  }
+
+  public NextAction handleFailure(StateExecution execution) {
+    if (handleFailureImpl(execution)) {
+      return moveToState(schedule, "Failure handled successfully");
+    }
+    return moveToState(failed, "Requires manual fixing");
```
efonsell: I wonder if we should have "ok" final state as well?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  protected boolean handleFailureImpl(StateExecution execution) {
+    logger.warn("Cron workflow " + getType() + " / " + execution.getWorkflowInstanceId() + "work failed");
```
efonsell:
```suggestion
    logger.warn("Cron workflow {} / {} work failed", getType(), execution.getWorkflowInstanceId());
```

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Cron workflow executor that periodically wakes up.
+ */
+public abstract class CronWorkflow extends WorkflowDefinition<State> {
+  private static final Logger logger = getLogger(CronWorkflow.class);
+
+  public static final String VAR_SCHEDULE = "cron";
+
+  public enum State implements WorkflowState {
+    schedule(start), doWork(normal), handleFailure(normal), failed(manual);
```
efonsell: is the doWork method intentionally left unimplemented?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,94 @@
...
+    setDescription("Does nflow maintenance periodically.");
+  }
+
+  public NextAction doWork(@SuppressWarnings("unused") StateExecution execution, @StateVar(value = VAR_MAINTENANCE_CONFIGURATON, readOnly = true) MaintenanceConfiguration conf) {
+    maintenanceService.cleanupWorkflows(conf);
+    return moveToState(schedule, "work done");
+  }
+
+  @Service
+  static class MaintenanceConfigurationService implements SmartLifecycle {
```
efonsell: extract to own file

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,94 @@
...
+  MaintenanceService maintenanceService;
+
+  public MaintenanceWorkflow() {
+    super(MAINTENANCE_WORKFLOW_TYPE);
+    setDescription("Does nflow maintenance periodically.");
+  }
+
+  public NextAction doWork(@SuppressWarnings("unused") StateExecution execution, @StateVar(value = VAR_MAINTENANCE_CONFIGURATON, readOnly = true) MaintenanceConfiguration conf) {
+    maintenanceService.cleanupWorkflows(conf);
+    return moveToState(schedule, "work done");
```
efonsell: I think this should be done in CronWorkflow

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,94 @@
...
+  static class MaintenanceConfigurationService implements SmartLifecycle {
+    private final MaintenanceConfiguration defaultConfiguration;
+    private final WorkflowInstanceService instanceService;
+    private final boolean insertOnStartup;
+
+    public MaintenanceConfigurationService(Environment env, WorkflowInstanceService instanceService) {
+      defaultConfiguration = new Builder()
+              .withDeleteArchivedWorkflows().setOlderThanPeriod(years(1)).done()
+              .withArchiveWorkflows().setOlderThanPeriod(days(2)).done()
+              .withDeleteWorkflows().setOlderThanPeriod(days(7)).done()
```
efonsell: there's nothing to delete older than 7 days in main tables if they are archived already after 2 days

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,94 @@
...
+              .build();
+      this.instanceService = instanceService;
+      this.insertOnStartup = env.getProperty("nflow.maintenance.auto_start", Boolean.class, false);
```
efonsell: _ or camelcase?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,94 @@
...
+      return defaultConfiguration;
+    }
+
+    @Override
+    public int getPhase() {
+      return MAX_VALUE;
+    }
+
+    @Override
+    public void start() {
```
efonsell: Use @EventListener(ApplicationReadyEvent.class) instead of implementing SmartLifecycle?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,94 @@
...
+      if (insertOnStartup) {
+        instanceService.insertWorkflowInstance(new WorkflowInstance.Builder()
+                .setType(MAINTENANCE_WORKFLOW_TYPE)
+                .putStateVariable(VAR_SCHEDULE, "4 4 4 * * *")
+                .putStateVariable(VAR_MAINTENANCE_CONFIGURATON, getDefaultConfiguration())
+                .setExternalId(MAINTENANCE_WORKFLOW_DEFAULT_EXTERNAL_ID)
+                .setNextActivation(now())
```
efonsell: this is not needed, it's activated now by default

nflow-rest-api-common/src/main/java/io/nflow/rest/v1/converter/MaintenanceConverter.java
```diff
@@ -17,24 +17,18 @@
...
-    }
+    ofNullable(request.archiveWorkflows).ifPresent(config -> convert(config, builder.withArchiveWorkflows()));
+    ofNullable(request.deleteArchivedWorkflows).ifPresent(config -> convert(config, builder.withDeleteArchivedWorkflows()));
+    ofNullable(request.deleteWorkflows).ifPresent(config -> convert(config, builder.withDeleteWorkflows()));
     return builder.build();
   }
 
-  private ConfigurationItem createConfig(MaintenanceRequestItem config) {
-    return new ConfigurationItem.Builder() //
+  private void convert(MaintenanceRequestItem config, MaintenanceConfiguration.ConfigurationItem.Builder builder) {
```
efonsell: convert sounds like it should return something, maybe rename to apply?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
gmokki: Yes. this is abstract workflow. That way it is possible to add the method with wanted signature (for example take in state variables)

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
gmokki: Does the workflow executor honor that or should the workflow itself also add code for it?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
gmokki: This is infinite workflow. So it should never and in an ok state.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
gmokki: Hmm. True. With the hierarchical archiving the deletion could have done a non-hierarchical delete for the children of long running workflows.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,88 @@
...
+
+  protected CronWorkflow(String type) {
+    this(type, new Builder().setHistoryDeletableAfter(weeks(1)).build());
+  }
+
+  public NextAction schedule(StateExecution execution, @StateVar(value = VAR_SCHEDULE, readOnly = true) String cron) {
+    return moveToStateAfter(doWork, nextActivationTime(cron, execution.getRequestedNextActivationTime()), "Scheduled");
+  }
+
+  protected DateTime nextActivationTime(String cron, DateTime activationTime) {
```
efonsell: nextActivationTime -> getNextActivationTime
DateTime activationTime -> DateTime previouslyRequestedActivationTime?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
efonsell: Ok, the action description is just funny and makes unnecessary assumptions. Maybe "Failure handling failed"?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,88 @@
...
+
+  public NextAction handleFailure(StateExecution execution) {
+    if (handleFailureImpl(execution)) {
+      return moveToState(schedule, "Failure handled successfully");
+    }
+    return moveToState(failed, "Requires manual fixing");
+  }
+
+  protected boolean handleFailureImpl(StateExecution execution) {
+    logger.warn("Cron workflow {} / {} work failed", getType(), execution.getWorkflowInstanceId());
```
efonsell: Should this be error? Maybe?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,37 @@
...
+import io.nflow.engine.workflow.definition.StateVar;
+import org.springframework.stereotype.Component;
+
+import javax.inject.Inject;
+
+import static io.nflow.engine.workflow.curated.CronWorkflow.State.schedule;
+import static io.nflow.engine.workflow.definition.NextAction.moveToState;
+
+/**
+ * Workflow that does nflow maintenance periodically.
```
efonsell:
```suggestion
 * Workflow that cleans up workflow instances periodically.
```

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+@Component
+public class MaintenanceWorkflow extends CronWorkflow {
+  public static final String MAINTENANCE_WORKFLOW_TYPE = "maintenance";
```
efonsell: Use a bit more specific type to avoid possible collisions.
```suggestion
  public static final String MAINTENANCE_WORKFLOW_TYPE = "nFlowMaintenance";
```
Could be "cleanupWorkflowInstances" also.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,37 @@
...
+  public static final String MAINTENANCE_WORKFLOW_DEFAULT_EXTERNAL_ID = "default";
+  public static final String VAR_MAINTENANCE_CONFIGURATION = "conf";
```
efonsell:
```suggestion
  public static final String VAR_MAINTENANCE_CONFIGURATION = "maintenanceConfig";
```

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Inject
+  MaintenanceService maintenanceService;
+
+  public MaintenanceWorkflow() {
+    super(MAINTENANCE_WORKFLOW_TYPE);
+    setDescription("Do nFlow maintenance periodically.");
```
efonsell:
```suggestion
    setDescription("Clean up workflow instances periodically.");
```

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  public NextAction doWork(@SuppressWarnings("unused") StateExecution execution, @StateVar(value = VAR_MAINTENANCE_CONFIGURATION, readOnly = true) MaintenanceConfiguration conf) {
+    maintenanceService.cleanupWorkflows(conf);
+    return moveToState(schedule, "work done");
```
efonsell: At least if the action text is given here, is should be more descriptive like "Workflow instances cleaned up". If the text is set in generic CronWorkflow, then it can be more generic "Work done"

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
```diff
@@ -0,0 +1,46 @@
...
+import static org.joda.time.Period.years;
+
+@Component
+public class MaintenanceWorkflowStarter {
+  protected final WorkflowInstanceService instanceService;
+  protected final boolean insertOnStartup;
+
+  public MaintenanceWorkflowStarter(Environment env, WorkflowInstanceService instanceService) {
+    this.instanceService = instanceService;
+    this.insertOnStartup = env.getProperty("nflow.maintenance.start", Boolean.class, false);
```
efonsell: I think the autoStart or auto_start was better. How about enabled? It might be nice to able to disable the workflow via config also after it has been inserted once. This would of course require that the workflow itself supports this property.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  protected String getDefaultCronSchedule() {
```
efonsell: rename to getCronSchedule (I suppose this can be overridden, and then it is not the default anymore)

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
```diff
@@ -0,0 +1,46 @@
...
+    return "4 4 4 * * *";
+  }
+
+  protected MaintenanceConfiguration getDefaultConfiguration() {
```
efonsell: rename to getMaintenanceConfiguration

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
```diff
@@ -0,0 +1,46 @@
...
+    return new MaintenanceConfiguration.Builder()
+            .withDeleteArchivedWorkflows().setOlderThanPeriod(years(1)).done()
+            .withArchiveWorkflows().setOlderThanPeriod(days(2)).done()
```
efonsell: This is pretty aggressive. Maybe a week or even month would be enough in most cases? At least in my project I use 90 days for archiving.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
gmokki: I used the term default since this only affects the initial time the workflow engine starts. After that the configuration is through the database

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
gmokki: What would be the harm in archiving aggressvively?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  @EventListener(ContextStartedEvent.class)
+  public void start() {
+    if (insertOnStartup) {
+      instanceService.insertWorkflowInstance(new WorkflowInstance.Builder()
+              .setType(MaintenanceWorkflow.MAINTENANCE_WORKFLOW_TYPE)
+              .putStateVariable(CronWorkflow.VAR_SCHEDULE, getDefaultCronSchedule())
+              .putStateVariable(MaintenanceWorkflow.VAR_MAINTENANCE_CONFIGURATION, getDefaultConfiguration())
+              .setExternalId(MaintenanceWorkflow.MAINTENANCE_WORKFLOW_DEFAULT_EXTERNAL_ID)
```
efonsell: If the constant is not static imported, maybe it could be renamed to EXTERNAL_ID? Even with static import, I would drop the DEFAULT from it. Maybe replace it by INSTANCE. Or something.

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/StateExecution.java
```diff
@@ -36,6 +37,14 @@
   int getRetries();
 
   /**
+   * Return the time when this workflow instance was originally scheduled to run.
+   * If there is much work to be done or executors are not running this can be far away in the past.
+   *
+   * @return The original next activation time of this workflow.
```
efonsell:
```suggestion
   * @return The original activation time of this workflow.
```

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/StateExecution.java
```diff
@@ -36,6 +37,14 @@
...
+   */
+  DateTime getRequestedNextActivationTime();
```
efonsell:
```suggestion
  DateTime getRequestedActivationTime();
```

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/StateExecution.java
```diff
@@ -36,6 +37,14 @@
   int getRetries();
 
   /**
+   * Return the time when this workflow instance was originally scheduled to run.
```
efonsell:
```suggestion
   * Return the time when this workflow instance was previously scheduled to run.
```

gmokki: It is the requested start time of the currently running step, not any previous step

nflow-rest-api-common/src/main/java/io/nflow/rest/v1/converter/MaintenanceConverter.java
```diff
@@ -17,24 +16,18 @@
...
-    if (request.archiveWorkflows != null) {
-      builder.setArchiveWorkflows(createConfig(request.archiveWorkflows));
-    }
-    if (request.deleteArchivedWorkflows != null) {
-      builder.setDeleteArchivedWorkflows(createConfig(request.deleteArchivedWorkflows));
-    }
-    if (request.deleteWorkflows != null) {
-      builder.setDeleteWorkflows(createConfig(request.deleteWorkflows));
-    }
+    ofNullable(request.archiveWorkflows).ifPresent(config -> apply(config, builder.withArchiveWorkflows()));
```
efonsell: public MaintenanceConfiguration convert(MaintenanceRequest request) {
    MaintenanceConfiguration.Builder builder = new MaintenanceConfiguration.Builder();
    apply(request.archiveWorkflows, () -> builder.withArchiveWorkflows());
    apply(request.deleteArchivedWorkflows, () -> builder.withDeleteArchivedWorkflows());
    apply(request.deleteWorkflows, () -> builder.withDeleteWorkflows());
    return builder.build();
  }

  private void apply(MaintenanceRequestItem requestItem, Supplier<Builder> builderSupplier) {
    ofNullable(requestItem).ifPresent(item -> builderSupplier.get() //
        .setOlderThanPeriod(item.olderThanPeriod) //
        .setBatchSize(item.batchSize) //
        .setWorkflowTypes(ofNullable(item.workflowTypes).orElse(emptySet())) //
        .done());
  }

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
efonsell: I see. Then it makes sense. Or should it be getInitialCronSchedule then?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
gmokki: I like the default since this is the default instance. If other instances are added it is advised to use non-default external id.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
efonsell: Makes investigating stuff a bit more difficult.

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/StateExecution.java
efonsell: Then "Return requested start time of the currently running step." And actually that time usually is set in the previous step.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
efonsell: Executor uses the error state given as constructor parameter unless it is overridden with permit.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
efonsell: I would go with protected abstract method to make the compiler complain about missing implementation, and moving the responsibility of returning the proper NextAction to CronWorkflow. But I can accept this as well.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,47 @@
...
+  MaintenanceService maintenanceService;
+
+  public MaintenanceWorkflow() {
+    super(MAINTENANCE_WORKFLOW_TYPE);
+    setDescription("Clean up workflow instances periodically.");
+  }
+
+  public NextAction doWork(@SuppressWarnings("unused") StateExecution execution, @StateVar(value = VAR_MAINTENANCE_CONFIGURATION, readOnly = true) MaintenanceConfiguration conf) {
+    MaintenanceResults results = maintenanceService.cleanupWorkflows(conf);
+    StringBuilder sb = new StringBuilder(64);
```
efonsell: This will look weird when there was nothing to do. Remove "Maintenance:" from the beginning?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
gmokki: Seehttps://github.com/NitorCreations/nflow/pull/381

efonsell: Does not really help when you need to investigate stuff directly from db.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,87 @@
...
+
+  protected CronWorkflow(String type) {
+    this(type, new Builder().setHistoryDeletableAfter(weeks(1)).build());
+  }
+
+  public NextAction schedule(StateExecution execution, @StateVar(value = VAR_SCHEDULE, readOnly = true) String cron) {
+    return moveToStateAfter(doWork, getNextActivationTime(cron), "Scheduled");
+  }
+
+  protected DateTime getNextActivationTime(String cron) {
```
efonsell: For overriding this and doing some other logic than cron, maybe pass the execution as parameter as well?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,52 @@
...
+@Component
+public class MaintenanceWorkflow extends CronWorkflow {
+  public static final String MAINTENANCE_WORKFLOW_TYPE = "nFlowMaintenance";
+  public static final String VAR_MAINTENANCE_CONFIGURATION = "config";
+
+  @Inject
+  MaintenanceService maintenanceService;
+
+  public MaintenanceWorkflow() {
+    super(MAINTENANCE_WORKFLOW_TYPE);
```
efonsell: It should be possible to override workflow settings of MaintenanceWorkflow

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
gmokki: With the normal scenario of multiple executors running the disable would need to be marked into the database (for example by clearing nextActivation). But it would be confusing that setting enable=false on one node would effect the others.
I now renamed this to _insertWorkflowIfMissing_.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
gmokki: Directly in db: it is as easy to search from archives than from main tables.
But I can later do a PR that provides a view that combines them

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
efonsell: True. Maybe the maintenance workflow could support a signal to skip the cleanup. If enabled, insert if not exists and clear the signal if it does. If not enabled, set the signal. Setting different values on different nodes would be strange anyway?

efonsell: Also: use getRequiredProperty and move the default to engine.properties

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
efonsell: Not really if the archive tables are huge and production tables are not. And this is not the only reason anyway. I will anyway need to override the default to much longer period in one of my projects, there's no need for more workarounds. I just think less aggressive defaults would be more safe. If the performance becomes a problem somewhere, it is super easy to configure more aggressive cleanup. Other way around, the damage has already been done.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
gmokki: I see no reason for anyone to ever extend MaintenanceWorkflow. It is much more cleaner to just extend CronWorkflow directly

efonsell: Overriding the handleFailure for one. Overriding the workflow settings for two. Everything else in MaintenanceWorkflow is reusable, I would not want to re-implement those parts. I tried, extending MaintenanceWorkflow was much more cleaner.

CHANGELOG.md
```diff
@@ -8,13 +8,16 @@
   - `StateExecution.getAllChildWorkflows` does not return current state variables of child workflows anymore.
   - `WorkflowStateProcessor` does not read child workflow IDs of processed workflow instances anymore.
   - `QueryWorkflowInstances.Builder` does not query for current state variables by default anymore.
+  - `BulkWorklow` moved to `curated` java package.
```
efonsell: full package name would be nice here

CHANGELOG.md
```diff
@@ -32,6 +35,7 @@
   - For example, `WorkflowSettings.setHistoryDeleteableAfterHours(12)` can now be achieved by `WorkflowSettings.setHistoryDeleteableAfter(Period.hours(12))`.
   - To get child workflows with state variables in a state method, use `StateExecution.queryChildWorkflows`.
   - To get all child workflow IDs in a state method, use `StateExecution.getAllChildWorkflows`.
+  - New `MaintenanceWorkflow` that can be enabled by setting `nflow.maintenance.insertWorkflowIfMissing` property to true. On the first run the workflow is inserted and it will start to run periodically. Further configuration must be done by editing the two state variables of the workflow: `cron` for the scheduling and `config` for the `MaintenanceConfiguration` to execute.
```
efonsell: Mention the config properties for initial config here

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,87 @@
...
+import static io.nflow.engine.workflow.definition.WorkflowStateType.normal;
+import static io.nflow.engine.workflow.definition.WorkflowStateType.start;
+import static org.joda.time.Instant.now;
+import static org.joda.time.Period.weeks;
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * Cron workflow executor that periodically wakes up.
+ */
+public abstract class CronWorkflow extends WorkflowDefinition<State> {
```
efonsell: Add javadocs for public and protected

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflow.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.inject.Inject;
+
+import static io.nflow.engine.workflow.curated.CronWorkflow.State.schedule;
+import static io.nflow.engine.workflow.definition.NextAction.moveToState;
+
+/**
+ * Workflow that cleans up workflow instances periodically.
+ */
+@Component
+public class MaintenanceWorkflow extends CronWorkflow {
```
efonsell: Add javadocs for public and protected

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/MaintenanceWorkflowStarter.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.function.Supplier;
+
+import static io.nflow.engine.workflow.curated.CronWorkflow.VAR_SCHEDULE;
+import static io.nflow.engine.workflow.curated.MaintenanceWorkflow.MAINTENANCE_WORKFLOW_TYPE;
+import static io.nflow.engine.workflow.curated.MaintenanceWorkflow.VAR_MAINTENANCE_CONFIGURATION;
+import static java.util.Optional.ofNullable;
+
+@Component
+public class MaintenanceWorkflowStarter {
```
efonsell: Javadocs

nflow-tests/src/test/java/io/nflow/tests/MaintenanceWorkflowTest.java
```diff
@@ -0,0 +1,111 @@
...
+  public void verifyThatMaintenanceWorkflowIsRunning() throws InterruptedException {
+    SECONDS.sleep(1);
+    ListWorkflowInstanceResponse[] instances = fromClient(workflowInstanceResource, true)
+            .query("type", MAINTENANCE_WORKFLOW_TYPE)
+            .query("externalId", MAINTENANCE_WORKFLOW_DEFAULT_EXTERNAL_ID)
+            .query("include", "currentStateVariables")
+            .get(ListWorkflowInstanceResponse[].class);
+    assertThat(asList(instances), hasSize(1));
+    assertThat(instances[0].stateVariables, hasEntry("cron", "* * * * * *"));
+//    assertThat(instances[0].stateVariables, hasEntry("config", hasEntry("deleteWorkflows", hasEntry("olderThanPeriod", "PT1S"))));
```
efonsell: should this be enabled or removed?

nflow-tests/src/test/java/io/nflow/tests/MaintenanceWorkflowTest.java
```diff
@@ -0,0 +1,111 @@
...
+  @Order(4)
+  public void waitForCleanup() throws InterruptedException {
+    SECONDS.sleep(6);
+  }
+
+  @Test
+  @Order(5)
+  public void verifyThatCleanupOccurred() {
+    ids.forEach(id ->
+            assertThrows(NotFoundException.class, () -> getWorkflowInstance(id))
```
efonsell: formatting looks odd

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,116 @@
...
+   * Extending class must implement the 'doWork` step method.
+   *
+   * @param type The type of the workflow.
+   */
+  protected CronWorkflow(String type) {
+    this(type, new Builder().setHistoryDeletableAfter(weeks(1)).build());
+  }
+
+  /**
+   * Calculates the next activation time.
```
efonsell: Default impl calcs next activation time based on cron state variable. Override for custom scheduling.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,116 @@
...
+   * @param execution The workflow execution context.
+   * @param cron The cron schedule.
+   * @return The next activation time.
+   */
+  protected DateTime getNextActivationTime(@SuppressWarnings("unused") StateExecution execution, String cron) {
+    return new DateTime(new CronSequenceGenerator(cron).next(now().toDate()));
+  }
+
+  /**
+   * Handles errors in working step.
```
efonsell: Default impl logs error and reschedules the workflow

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * @param execution The workflow execution context.
+   * @return True if the failure was handled successfully. If false is returned the cron workflow stops waiting for manual actions.
+   */
+  protected boolean handleFailureImpl(StateExecution execution) {
+    logger.error("Cron workflow {} / {} work failed", getType(), execution.getWorkflowInstanceId());
+    return true;
+  }
+
+  public NextAction schedule(StateExecution execution, @StateVar(value = VAR_SCHEDULE, readOnly = true) String cron) {
```
efonsell: Javadoc

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,116 @@
...
+    return moveToStateAfter(doWork, getNextActivationTime(execution, cron), "Scheduled");
+  }
+
+  public NextAction handleFailure(StateExecution execution) {
```
efonsell: Javadoc. Could we actually let the extending class override this state method if custom handling is needed? E.g. just move handledFailureImpl default impl here?

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/TestCronWorkflow.java
```diff
@@ -0,0 +1,21 @@
...
+import io.nflow.engine.workflow.curated.CronWorkflow;
+import io.nflow.engine.workflow.definition.NextAction;
+import io.nflow.engine.workflow.definition.StateExecution;
+import org.springframework.stereotype.Component;
+
+import static io.nflow.engine.workflow.curated.CronWorkflow.State.schedule;
+import static io.nflow.engine.workflow.definition.NextAction.moveToState;
+
+@Component
+public class TestCronWorkflow extends CronWorkflow {
```
efonsell: It would be nice to show here all the things that can be overridden (handleFailure, settings, getNextActivation)

efonsell: Or maybe one SimpleCron example and then another one for showing all the moving parts

nflow-tests/src/test/java/io/nflow/tests/MaintenanceTest.java
```diff
@@ -163,15 +161,23 @@ private long createWorkflow() {
...
-  private int archiveOlderThan(DateTime olderThan) {
+  private int archiveOlderThan(DateTime now) {
+    return archiveOlderThan(maintenanceResource, now);
+  }
+
+  private int deleteOlderThan(DateTime now) {
+    return deleteOlderThan(maintenanceResource, now);
+  }
+
+  static int archiveOlderThan(WebClient maintenanceResource, DateTime olderThan) {
```
efonsell: Why passing maintenanceResource as parameter and having these static helpers?

efonsell: Ah it is used in another test. Move helpers to AbstractNflowTest instead?

nflow-tests/src/test/java/io/nflow/tests/extension/NflowServerExtension.java
```diff
@@ -1,46 +1,77 @@
...
+    }
+
     @Override
     public void afterAll(ExtensionContext context) {
+        invokeBeforeServerStop();
         config.after();
     }
 
+    private void invokeBeforeServerStop() {
+        Arrays.stream(testClass.getMethods())
```
efonsell: static import

nflow-tests/src/test/java/io/nflow/tests/extension/NflowServerExtension.java
```diff
@@ -1,46 +1,77 @@
...
+                .filter(field -> field.getAnnotation(BeforeServerStop.class) != null)
+                .forEach(method -> {
+            try {
```
efonsell: extract try-catch to method

efonsell:
Please remove unused imports

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/TestCronWorkflow.java
gmokki: This just tests that the cron workflow works. I think we have examples separately?

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,132 @@
...
+   */
+  public NextAction schedule(StateExecution execution, @StateVar(value = VAR_SCHEDULE, readOnly = true) String cron) {
+    return moveToStateAfter(doWork, getNextActivationTime(execution, cron), "Scheduled");
+  }
+
+  /**
+   * Tries to handle failures by calling {@link #handleFailureImpl}
+   *
+   * @param execution The workflow execution context.
+   * @return Either moves back to schedule state or if failure handling failed waits for manual actions in failed state.
```
efonsell:
```suggestion
   * @return Action to either move back to schedule state or to move to failed state to wait for manual actions.
```

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,132 @@
...
+   */
+  public NextAction handleFailure(StateExecution execution) {
+    if (handleFailureImpl(execution)) {
+      return moveToState(schedule, "Failure handled successfully");
```
efonsell: We don't know if the failure was handled successfully, it is up to handleFailureImpl to decide how to continue regardless of the failure handling status.
```suggestion
      return moveToState(schedule, "Failure handled, rescheduling");
```

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+    return moveToState(failed, "Failure handling failed");
```
efonsell:
```suggestion
    return moveToState(failed, "Failure handled, moving to failed state");
```

nflow-tests/src/main/java/io/nflow/tests/demo/workflow/TestCronWorkflow.java
efonsell: I think the nflow-tests package is the best place for working examples as they are continuously tested and thus easy to keep up to date when things change.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
efonsell: nFlow engine did not like that:
```
      if (methods.containsKey(method.getName())) {
        throw new IllegalStateException("Method " + method + " was overloaded. Overloading state methods is not allowed.");
      }
```

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/WorkflowSettings.java
```diff
@@ -94,7 +97,36 @@
...
+    public static BooleanSupplier onAverageEveryNthExecution(int n) {
+      return () -> ThreadLocalRandom.current().nextInt(n) == 0;
+    }
+
+    /**
+     * Returns true randomly once per day (during the early hours).
+     *
+     * @return Producer of boolean values
+     */
+    public static BooleanSupplier oncePerDay() {
```
efonsell: There could be a test for this

PR #382, opened 2020-02-20
Stream results out of rest api. 

gmokki:
The main table query is still done synchronously, but additional queries are done on the stream lowering latency.
Should help with explorer UI, especially if it is later updated to process results lines as they arrive.


jsyrjala:
Looks good to me. 

I think this will not actually speed up UI that much, since it will wait until it has received the whole JSON before showing anything.

PR #386, opened 2020-03-17
Support for moving to a failure state without retries based on thrown exception and state

efonsell:


nflow-engine/src/main/java/io/nflow/engine/workflow/definition/WorkflowSettings.java
```diff
@@ -68,6 +69,12 @@
    * Default priority for new workflow instances.
    */
   public final short defaultPriority;
+  /**
+   * Checks if the state method should be retried after an exception was thrown by it. By default, only exceptions that are
+   * annotated with {@code @NonRetryableError} will cause immediate transition to the failure state of the executed state, all
+   * others are considered retryable.
+   */
+  public final Function<Throwable, Boolean> isRetryable;
```
gmokki: Maybe this is better as `Predicate<Throwable>` ?

efonsell: This was completely removed

PR #388, opened 2020-04-05
Tune sql

gmokki:
SQL tuning


efonsell:
Will there be upgrade scripts?

gmokki:
Yes. But only after we decide which commits are good idea. Like was the order of the uniq index fields (type before external_id) really tested to help for some cases. I think it is bad for then engine code, but maybe ui uses it?
And should we even add any extra archive indices (like type), since we have not tested if the db actually uses them (since we have usually order by id in all queries)

nflow-engine/src/main/resources/scripts/db/update-7.1.0-x/postgresql.update.ddl.sql
```diff
@@ -0,0 +1,25 @@
...
+drop index idx_workflow_archive_parent;
+create index concurrently idx_workflow_archive_parent on nflow_archive_workflow(parent_workflow_id) where parent_workflow_id is not null;
+
+alter table nflow_workflow set (fillfactor=95);
+
+alter index nflow_workflow_action_pkey set (fillfactor=100);
+alter index nflow_workflow_action_workflow set (fillfactor=100);
+alter index pk_workflow_state set (fillfactor=100);
+
+drop index nflow_archive_workflow_uniq;
```
efonsell, line 13: Why this is removed?

nflow-engine/src/main/resources/scripts/db/update-7.1.0-x/postgresql.update.ddl.sql
```diff
@@ -0,0 +1,25 @@
...
+alter index idx_workflow_archive_parent set (fillfactor=100);
+alter index nflow_archive_workflow_action_pkey set (fillfactor=100);
+alter index nflow_archive_workflow_action_workflow set (fillfactor=100);
+alter index pk_arch_workflow_state set (fillfactor=100);
+
+alter table nflow_worklow rename constraint nflow_workflow_uniq to nflow_workflow_uniq_old;
+alter table nflow_worklow add constraint nflow_workflow_uniq unique (external_id, type, executor_group);
+alter table nflow_worklow drop constraint nflow_workflow_uniq_old;
+
+create index idx_workflow_archive_type on nflow_archive_workflow(type) with (fillfactor=100);
```
efonsell, line 25: No newline at end of file

nflow-engine/src/main/resources/scripts/db/update-7.1.0-x/postgresql.update.ddl.sql
gmokki: The uniqueness is no longer important in archive. It is meant for the to support idempotent insert of workflow instances.
So I was thinking that archive might actually benefit more from index on type, which I added instead.

gmokki: But maybe I should drop that archive index from all databases?

efonsell: I'm not sure, maybe it's good it's dropped, then you can add new instance to prod with same extid and type, and archive that even though there is already such item in archive. It might be good idea to document this somewhere where we say that type+extid is unique. At least highlight this is release notes.
And yes, if it's dropped, it should be dropped from all databases.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
   - Throw `IllegalArgumentException` instead of `IllegalStateException` when trying to update workflow instance state to an invalid value.
   - Throw `IllegalArugmentException` instead of `RuntimeException` when trying to insert workflow instance with unknown type or with a state that is not a start state.
   - Make `StateVariableTooLongException` extend `IllegalArgumentException` instead of `RuntimeException`.
+  - MySQL/MariaDB: use compressed table format for archives.
+  - PostgreSQL/SQLServer: use partial (not null) indices for parent hierarchy
```
efonsell, line 16:
```suggestion
  - PostgreSQL/SQLServer: use partial (not null) indices for parent hierarchy.
```

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
...
+  - PostgreSQL: tune fillfactors of tables and indices
```
efonsell, line 17:
```suggestion
  - PostgreSQL: tune fillfactors of tables and indices.
```

nflow-engine/src/main/resources/scripts/db/sqlserver.create.ddl.sql
```diff
@@ -131,11 +131,8 @@ create table nflow_archive_workflow (
   constraint nflow_archive_workflow_uniq unique (type, external_id, executor_group)
 );
 
-if not exists (select 1 from sys.indexes where name='nflow_archive_workflow_parent')
-create index nflow_archive_workflow_parent on nflow_archive_workflow(parent_workflow_id, parent_action_id);
```
efonsell, line 135: Remove this from upgrade script as well

nflow-engine/src/main/resources/scripts/db/update-7.1.0-x/postgresql.update.ddl.sql
efonsell: Let's drop this from all databases.

nflow-engine/src/main/resources/scripts/db/update-7.1.0-x/sqlserver.update.ddl.sql
```diff
@@ -1,5 +1,7 @@
 drop index nflow_archive_workflow_uniq;
 
+drop index nflow_archive_workflow_parent;
```
gmokki, line 3: If we drop the index on update, should we also remove the index creation from the create `sqlserver.create.ddl.sql`?

PR #390, opened 2020-04-06
Better error respones from WorkflowInstanceResource endpoints

efonsell:
Change nflow-engine to throw IllegalArgumentExceptions instead of IllegalStateExceptions/RuntimeExceptions
Construct error responses with proper HTTP code (400 instead of 500) and message text in WorkflowInstanceResources
Remove ExceptionMapper from nflow-jetty for these cases


nflow-tests/src/test/java/io/nflow/tests/StateVariablesTest.java
```diff
@@ -109,7 +108,7 @@ public void updateWorkflowWithTooLongStateVariableValueReturnsBadRequest() {
 
     try (Response response = getInstanceIdResource(createResponse.id).put(req)) {
       assertThat(response.getStatus(), is(BAD_REQUEST.getStatusCode()));
-      assertThat(response.readEntity(ErrorResponse.class).error, startsWith("Too long value"));
+      assertThat(response.readEntity(String.class), startsWith("Too long value"));
```
gmokki, line 111: We no longer return json responses? I think this is a bad change.

gmokki: Also verify that response content type is application/json

nflow-rest-api-spring-web/src/main/java/io/nflow/rest/v1/springweb/WorkflowInstanceResource.java
```diff
@@ -81,12 +83,16 @@ public WorkflowInstanceResource(WorkflowInstanceService workflowInstances, Creat
...
     WorkflowInstance instance = createWorkflowConverter.convert(req);
-    long id = workflowInstances.insertWorkflowInstance(instance);
-    instance = workflowInstances.getWorkflowInstance(id, EnumSet.of(WorkflowInstanceInclude.CURRENT_STATE_VARIABLES), null);
-    return created(URI.create(String.valueOf(id))).body(createWorkflowConverter.convert(instance));
+    try {
+      long id = workflowInstances.insertWorkflowInstance(instance);
+      instance = workflowInstances.getWorkflowInstance(id, EnumSet.of(WorkflowInstanceInclude.CURRENT_STATE_VARIABLES), null);
+      return created(URI.create(String.valueOf(id))).body(createWorkflowConverter.convert(instance));
+    } catch (IllegalArgumentException e) {
+      return status(BAD_REQUEST).body(e.getMessage());
```
gmokki, line 94: We promise the produce json output. Plain string is not valid json.

nflow-jetty/src/main/java/io/nflow/jetty/mapper/StateVariableValueTooLongExceptionMapper.java
```diff
@@ -1,18 +0,0 @@
...
-
-import javax.ws.rs.core.Response;
-import javax.ws.rs.ext.ExceptionMapper;
-import javax.ws.rs.ext.Provider;
-
-import io.nflow.engine.workflow.executor.StateVariableValueTooLongException;
-import io.nflow.rest.v1.msg.ErrorResponse;
-
-@Provider
-public class StateVariableValueTooLongExceptionMapper implements ExceptionMapper<StateVariableValueTooLongException> {
```
gmokki, line 13: I think there is a tradeoff between putting the error catching&converting logic to one place vs replicating it in every method of the api.
I think we could have one generic exception mapper that just catches Throwable and returns ErrorResponse object.

PR #392, opened 2020-04-16
Guice support

efonsell:


nflow-engine/src/main/java/io/nflow/engine/internal/guice/EngineModule.java
```diff
@@ -0,0 +1,212 @@
...
+  @Singleton
+  @Inject
+  public SQLVariants nflowSQLVariants(Environment env, DatabaseConfiguration databaseConfiguration) {
+    return databaseConfiguration.sqlVariants(env);
+  }
+
+  @Provides
+  @Singleton
+  @Inject
+  public DatabaseConfiguration databaseConfiguration(Environment env) {
```
gmokki, line 157: Maybe take `Collection<Class<SQLVariants>>` as parameter? and then loop through to find a match for the env?

efonsell: Decided to keep it like this

PR #393, opened 2020-04-17
Inform workflow dispatcher about pending shutdown to allw faster clean shutdown

gmokki:


nflow-engine/src/main/java/io/nflow/engine/internal/executor/WorkflowStateProcessor.java
```diff
@@ -267,15 +269,8 @@ private WorkflowInstance saveWorkflowInstanceState(StateExecutionImpl execution,
...
-    } while (internalRetryEnabled);
-    throw new IllegalStateException(format("Failed to save workflow instance %s new state", instance.id));
-  }
-
-  /**
-   * For unit testing only
-   */
-  void setInternalRetryEnabled(boolean internalRetryEnabled) {
-    this.internalRetryEnabled = internalRetryEnabled;
+    } while (!shutdownRequested.get());
```
efonsell, line 272: Check shutdown in catch block above. If shutting down, log error (failed, shutting down) and return the instance. If not, log error (failed, retrying) and go to sleep.
And check for shutdown in run method catch block also. If shutting down, log error (failed, shutting down) and stateProcessingFinished = true, else log error (failed, retrying) and got to sleep.

nflow-engine/src/test/java/io/nflow/engine/internal/executor/WorkflowStateProcessorFactoryTest.java
```diff
@@ -61,14 +62,14 @@ public void setup() {
 
   @Test
   public void factoryCreatesExecutorsWithoutListeners() {
-    WorkflowStateProcessor executor = factory.createProcessor(12);
+    WorkflowStateProcessor executor = factory.createProcessor(12, () -> FALSE);
```
efonsell, line 65: Use either FALSE::booleanValue or () -> FALSE consistently?

nflow-engine/src/test/java/io/nflow/engine/internal/executor/WorkflowStateProcessorTest.java
```diff
@@ -826,14 +834,30 @@ public void saveStateRetryAfterFailedPersistence() throws InterruptedException {
...
+
+    ExecutorService executorService = newSingleThreadExecutor();
+    executorService.submit(executor);
+    sleep(500);
+
+    executorService.shutdown();
+    shutdownRequest.set(true);
+    executorService.awaitTermination(10, SECONDS);
+
+    assertThat(loopingWf.counter, Matchers.lessThan(20));
```
efonsell, line 858: static import

nflow-engine/src/test/java/io/nflow/engine/internal/executor/WorkflowStateProcessorTest.java
```diff
@@ -1078,6 +1102,38 @@ public NextAction start(@SuppressWarnings("unused") StateExecution execution) {
...
+
+      @Override
+      public WorkflowStateType getType() {
+        return stateType;
+      }
+    }
+
+    public NextAction start(@SuppressWarnings("unused") StateExecution execution) throws InterruptedException {
+      sleep(100);
+      return counter++ < 100 ? moveToState(State.start, "loop") : moveToState(State.done, "done");
```
efonsell, line 1132: test will fail anyway if looping more than 20 times, wait delay the failure for 80 iterations after that?

gmokki: just to make extra sure some random timing/slowness problem on travis does not break the test. When the test works it takes 1.5 seconds so how long it takes if the code breaks does not matter. At least it is clear when it takes 10 seconds that it was not bad luck.

efonsell: After the counter has reached 20, the test will fail anyway regardless of possible timing issues because of `assertThat(loopingWf.counter, lessThan(20));`
I would actually replace that assert with a check that awaitTermination(5, seconds) returns true (e.g. executor terminated, 5 second timeout should be enough there) and workflow state is `start` (termination was due to shutdown request, not because workflow reached done state).

PR #396, opened 2020-04-20
Map exceptions to error responses in all rest endpoints

efonsell:


nflow-rest-api-jax-rs/src/main/java/io/nflow/rest/v1/jaxrs/WorkflowInstanceResource.java
```diff
@@ -116,12 +109,10 @@ public Response createWorkflowInstance(
     @ApiResponse(code = 409, message = "If workflow was executing and no update was done") })
   public Response updateWorkflowInstance(@ApiParam("Internal id for workflow instance") @PathParam("id") long id,
       @ApiParam("Submitted workflow instance information") UpdateWorkflowInstanceRequest req) {
-    try {
+    return handleExceptions(() -> {
```
gmokki, line 112: Can this throw EmptyResultDataAccessException if the workflow is not found?
-> Your exception mapper will then fail to format the string due to missing argument?

efonsell: It may, but not always. Anyway the exception mapper does not fail on this anymore.

nflow-engine/src/main/java/io/nflow/engine/service/WorkflowInstanceService.java
```diff
@@ -90,23 +95,29 @@ public boolean updateWorkflowInstance(WorkflowInstance instance, WorkflowInstanc
...
+      }
+      WorkflowInstance updatedInstance = builder.build();
+      boolean updated = workflowInstanceDao.updateNotRunningWorkflowInstance(updatedInstance);
+      if (updated) {
+        String currentState = workflowInstanceDao.getWorkflowInstanceState(updatedInstance.id);
+        WorkflowInstanceAction updatedAction = new WorkflowInstanceAction.Builder(action).setState(currentState).build();
+        workflowInstanceDao.insertWorkflowInstanceAction(updatedInstance, updatedAction);
+      } else {
+        // this is to trigger EmptyResultDataAccessException if instance does not exist
+        workflowInstanceDao.getWorkflowInstance(instance.id, emptySet(), 0L);
```
gmokki, line 115: Why not just `throw NflowNotFoundException("Workflow instance", instance.id, e)` directly?

nflow-rest-api-spring-web/src/main/java/io/nflow/rest/v1/springweb/SpringWebResource.java
```diff
@@ -0,0 +1,34 @@
...
+    } catch (IllegalArgumentException e) {
+      return toErrorResponse(BAD_REQUEST, e.getMessage());
+    } catch (NflowNotFoundException e) {
+      return toErrorResponse(NOT_FOUND, e.getMessage());
+    } catch (Throwable t) {
+      return toErrorResponse(INTERNAL_SERVER_ERROR, t.getMessage());
+    }
+  }
+
+  private ResponseEntity<?> toErrorResponse(HttpStatus status, String error) {
```
gmokki, line 31: Only this method is different between sprinweb and jaxrs resource. Maybe move the handleException to ResourceBase and make toErrorResponse abstract there?

gmokki: You would need to make ResourceBase generic so that you pass the Response object class there type parameter.

nflow-engine/src/main/java/io/nflow/engine/service/WorkflowInstanceService.java
efonsell: update can be false also when instance exists (but is executing and thus not updated)

nflow-rest-api-spring-web/src/main/java/io/nflow/rest/v1/springweb/SpringWebResource.java
efonsell: I was thinking about that, but also the statuses are different (HttpStatus vs Response.Status). But if I use ints instead of those, maybe it works. I'll give it a try.

PR #397, opened 2020-04-20
Use for update skip locked for postgresql polling

gmokki:
Fixes a deadlock observed when multiple engines are running concurrently

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
   - Throw `IllegalArgumentException` instead of `IllegalStateException` when trying to update workflow instance state to an invalid value.
   - Throw `IllegalArugmentException` instead of `RuntimeException` when trying to insert workflow instance with unknown type or with a state that is not a start state.
   - Make `StateVariableTooLongException` extend `IllegalArgumentException` instead of `RuntimeException`.
+  - Fix sql deadlocks in polling for PostgreSQL with skip locked
   - Dependency updates:
     - spring 5.2.5
     - jackson 2.10.3
--- a/nflow-engine/src/main/java/io/nflow/engine/config/db/SqlServerDatabaseConfiguration.java
+++ b/nflow-engine/src/main/java/io/nflow/engine/config/db/SqlServerDatabaseConfiguration.java
@@ -37,6 +37,7 @@ public class SqlServerDatabaseConfiguration extends DatabaseConfiguration {
 
   /**
    * Creates the SQL variants for SQL Server database.
+   *
    * @return SQL variants optimized for SQL Server.
    */
   @Bean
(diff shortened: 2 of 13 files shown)


nflow-engine/src/main/java/io/nflow/engine/config/db/SqlServerDatabaseConfiguration.java
```diff
@@ -97,8 +97,8 @@ public boolean hasUpdateableCTE() {
     }
 
     @Override
-    public String forUpdateInnerSelect() {
-      return "";
+    public String forUpdateInnerSelect(boolean skipLocked) {
+        return skipLocked ? " with (updlock,readpast)" : " with (updlock)";
```
jsyrjala, line 101: This is always called with `true`. Is this conditional needed?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
   - Throw `IllegalArgumentException` instead of `IllegalStateException` when trying to update workflow instance state to an invalid value.
   - Throw `IllegalArugmentException` instead of `RuntimeException` when trying to insert workflow instance with unknown type or with a state that is not a start state.
   - Make `StateVariableTooLongException` extend `IllegalArgumentException` instead of `RuntimeException`.
+  - Fix sql deadlocks in polling for PostgreSQL with skip locked
```
efonsell, line 15: Seems like this is done for all databases? sql -> SQL

nflow-tests/src/test/java/io/nflow/tests/ConcurrentEnginesTest.java
```diff
@@ -0,0 +1,124 @@
...
+                .query("businessKey", UNIQUE_KEY)
+                .query("status", finished)
+                .query("include", "").get(ListWorkflowInstanceResponse[].class);
+        if (instances.length == WORKFLOWS) {
+          return instances;
+        }
+      }
+    });
+    assertThat(wfr.length, is(WORKFLOWS));
+    List<DateTime> startTimes = Stream.of(wfr).map(r -> r.started).sorted().collect(toList());
```
efonsell, line 121: startTimes is not used, can this be removed?

nflow-engine/src/main/java/io/nflow/engine/config/db/SqlServerDatabaseConfiguration.java
gmokki: I think it is more clear this way. But I could rename the method to forUpdateInnerSelectSkipLocked and remove the boolean. Not sure if we ever need the false option

CHANGELOG.md
gmokki: Other databases had a completely different fix already few releases ago. This code path is only for pg

nflow-tests/src/test/java/io/nflow/tests/ConcurrentEnginesTest.java
gmokki: I was thinking of coming up with something to test with the start times, comparing the max between avg. But could not get any difference with or without deadlocks. What would be a good way to test if the deadlock condition occured?

nflow-engine/src/main/java/io/nflow/engine/config/db/SqlServerDatabaseConfiguration.java
efonsell: IMO remove boolean and rename if needed

CHANGELOG.md
efonsell: Why there are changes in SqlServerDatabaseConfiguration then?

nflow-tests/src/test/java/io/nflow/tests/ConcurrentEnginesTest.java
efonsell: If there's no easy way to test it, just remove it?

CHANGELOG.md
gmokki: The SQL server does support skip locked, and it uses it now in archiving code. But there should never have been any deadlocks there (or at least the archived rows should be so old as to not conflict with engine). The only database that uses the `select * from workflow where id in (update set executor where executor is null returning id)` is postgresql. it is the only sql variant that currently returns true for `hasUpdateReturning`. Technically at least ms sqlserver supports it too, but syntax is very different with `OUTPUT` so I have not taken it into use yet.

PR #415, opened 2020-12-21
fix sql perf / memory issue

efonsell:


nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -698,13 +696,22 @@ private long getMaxResults(Long maxResults) {
...
+    if (includeStateVariables) {
+      Map<Long, Map<String, String>> actionStates = fetchActionStateVariables(instance, actionBuilders.size(), maxActions);
+      actionBuilders.forEach(builder -> {
+        Map<String, String> actionState = actionStates.get(builder.getId());
+        if (actionState != null) {
+          builder.setUpdatedStateVariables(actionState);
+        }
+      });
+    }
+    instance.actions.addAll(actionBuilders.stream().map(WorkflowInstanceAction.Builder::build).collect(toList()));
```
gmokki, line 714: `actionBuilders.stream().map(WorkflowInstanceAction.Builder::build).forEach(instance.actions::add)`

gmokki (approved):
Should we also add a test that creates wf with 10 actions and queries only 5 of them with state variables?

gmokki (approved):

PR #421, opened 2021-01-22
Search workflow instances by state variable key and value

efonsell:


nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -628,46 +628,56 @@ public int compareTo(OptimisticLockKey other) {
...
-    conditions.add("executor_group = :executor_group");
+    conditions.add("wf.executor_group = :executor_group");
     params.addValue("executor_group", executorInfo.getExecutorGroup());
-    sql += " where " + collectionToDelimitedString(conditions, " and ") + " order by id desc";
+    if (query.stateVariableKey != null) {
+      sqlBuilder.append("inner join nflow_workflow_state outside on wf.id = outside.workflow_id ")
+          .append("inner join (select workflow_id, max(action_id) action_id from nflow_workflow_state ")
+          .append("where state_key = :state_key group by workflow_id) inside ")
+          .append("on outside.workflow_id = inside.workflow_id and outside.action_id = inside.action_id ");
+      conditions.add("outside.state_key = :state_key");
```
gmokki, line 675: I've been thinking that we could try also sql that looks like
```sql
   inner join nflow_workflow_state outside on wf.id = outside.workflow_id and outside.state_key = :state_key and outside.state_value = :state_value
```
```java
  conditions.add("outside.action_id = (select max(action_id) from nflow_workflow_state where workflow_id = wf.id and state_key = :state_key)"); 
```

gmokki:
And I think we do need one test in the testa package that uses the rest api to actually fetch a workflow. Maybe add something to statevar test or some other?

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
```diff
@@ -628,46 +628,56 @@ public int compareTo(OptimisticLockKey other) {
...
 
   public Stream<WorkflowInstance> queryWorkflowInstancesAsStream(QueryWorkflowInstances query) {
-    String sql = "select * from nflow_workflow ";
+    StringBuilder sqlBuilder = new StringBuilder("select wf.* from nflow_workflow wf ");
     List<String> conditions = new ArrayList<>();
     MapSqlParameterSource params = new MapSqlParameterSource();
     conditions.add(executorInfo.getExecutorGroupCondition());
     if (!isEmpty(query.ids)) {
-      conditions.add("id in (:ids)");
+      conditions.add("wf.id in (:ids)");
```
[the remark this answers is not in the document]
efonsell: I thinks it's good as there are multiple tables it could refer to

efonsell:
> And I think we do need one test in the testa package that uses the rest api to actually fetch a workflow. Maybe add something to statevar test or some other?

Yes, I will add a test for this

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
efonsell: Yeah, we could try that also

gmokki (approved):

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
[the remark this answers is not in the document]
efonsell: No, you were right, it's unnecessary. Removed.

gmokki (approved):

nflow-engine/src/main/java/io/nflow/engine/internal/dao/WorkflowInstanceDao.java
efonsell: Changed

PR #430, opened 2021-02-05
Minor explorer tuning

eputtone:
1. Add hooks for customizing Explorer look through config.js (header logo, hide footer, change HTML title)
2. New look for workflow instance properties (previous made your eyes bleed). Also changed the order a bit and removed redundant workflow instance id (it was shown after workflow definion name already)
3. Layout of workflow instance properties and graph changed. Usually everything (properties, graph, tabs) fit "on the same line", but for large workflows (>15 states) rendering remains as it was (graph below workflow instance properties)
4. Separate ugly link from workflow instance page removed and workflow type name converted to a link.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 **Highlights**
 - Support updating workflow instance business key.
 - Support for searching workflow instances by state variable key and value.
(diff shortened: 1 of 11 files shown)


efonsell:
Changelog?

PR #432, opened 2021-02-11
Workflow definition refactoring

efonsell:


nflow-engine/src/main/java/io/nflow/engine/workflow/curated/SimpleState.java
```diff
@@ -0,0 +1,45 @@
+package io.nflow.engine.workflow.curated;
+
+import io.nflow.engine.model.ModelObject;
+import io.nflow.engine.workflow.definition.WorkflowState;
+import io.nflow.engine.workflow.definition.WorkflowStateType;
+
+public class SimpleState extends ModelObject implements WorkflowState {
```
eputtone, line 7: Add some JavaDoc to describe what this curated class is? I suppose this is "the default implementation", which would be named `WorkflowState` if the interface had not stolen the name already? And `WorkflowStateImpl` sounds archaic.

Still, `SimpleState` makes me look for `ComplexState` or alternative state implementations. I expect customizing state implementation will be pretty rare. Naming is difficult, and I can go forward with `SimpleState` too, but right now e.g. `BaseState` would sound better - communicates a possible extension point and makes you less surprised there's no offered alternatives.

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/AbstractWorkflowDefinition.java
```diff
@@ -101,23 +105,33 @@ public String getType() {
...
-  public abstract Set<S> getStates();
+  public Set<WorkflowState> getStates() {
+    return new HashSet<>(states);
+  }
+
+  /**
+   * Register a state to the workflow.
+   * @param state The state to be registered.
+   */
+  public void registerState(WorkflowState state) {
```
eputtone, line 132: If I understood correctly - earlier the state space was defined as enum values. This assured in compile-time that you cannot make non-existent state transitions.

After this PR there's a requirement that you explicitly register the states. `permit()` does this as a side-effect, but if you earlier had a state without incoming transitions (=no permits) then you need to add `registerState()` call for those orphan states when you migrate to 8.0.0. This could be mentioned in `CHANGELOG` and perhaps in the JavaDoc of this class too. E.g. `permit()` javadoc could say it does the registration (instead of "adding", whatever that means), and class-level javadoc that you need to register all the states now.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/SimpleState.java
efonsell: Renamed `SimpleState` -> `State` and added javadoc

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/AbstractWorkflowDefinition.java
efonsell: AFAIK the compile-time safety is not really changed here. Firstly, when using `WorkflowInstanceService.updateWorkflowInstance`, the state was defined as `string`. Secondly, when using `NextAction.moveToState` etc., the state was defined as `WorkflowState`. So it was possible to make non-existent state transitions before as well. IMO it should be the responsibility of unit tests to ensure that state transitions are correct.

Regarding the new requirement to register all states, I updated the javadocs and changelog.

nflow-engine/src/main/java/io/nflow/engine/workflow/curated/SimpleState.java
eputtone: Great! I approve :)

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/AbstractWorkflowDefinition.java
eputtone: Looks good! Meaning auto-scanning static states etc. I can't find ways to improve it further here.

eputtone (approved):
Good job! Looking more like 2020's now.

CHANGELOG.md
```diff
@@ -1,11 +1,18 @@
...
 
 **Details**
 - `nflow-engine`
-  - Dependency updates:
-    - minimum supported maven version for buliding is 3.8
+  - Workflow definitions that used to extend `WorkflowDefinition` should now extend `AbstractWorkflowDefinition` instead.
+    - It is not necessary to define the workflow states as an enum anymore, which should make it easier to extend and reuse states across different workflow definitions.
+    - You can define the states as instances of `io.nflow.engine.workflow.curated.State` or anything else that implements the required `io.nflow.engine.workflow.definition.WorkflowState` interface.
+    - The workflow definitions must now register all possible states as described in `io.nflow.engine.workflow.definition.AbstractWorkflowDefinition`.
+  - `WorkflowState.isRetryAllowed` was removed, use `new WorkflowSettings.Builder().setExceptionAnalyzer(...)` instead.
```
gmokki, line 14: Is it possible to provide a ready-made exception analyzer class that has the same functionality?

nflow-engine/src/main/java/io/nflow/engine/internal/workflow/WorkflowDefinitionScanner.java
```diff
@@ -91,6 +94,15 @@
     return methods;
   }
 
+  public Set<WorkflowState> getStaticWorkflowStates(Class<?> definition) {
+    final Set<WorkflowState> states = new HashSet<>();
+    doWithFields(definition, field -> {
+      field.setAccessible(true);
```
gmokki, line 100: I would add try-catch around the set-accessible, or alternatively only run it if the field is not already accessible.
The newer javas are more strict on things.

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/AbstractWorkflowDefinition.java
```diff
@@ -18,47 +19,102 @@
...
     if (stateMethods != null) {
       this.stateMethods = stateMethods;
     } else {
-      this.stateMethods = new WorkflowDefinitionScanner().getStateMethods(getClass());
+      this.stateMethods = scanner.getStateMethods(getClass());
     }
-    requireStateMethodExists(initialState);
+    registerState(initialState);
+    registerState(errorState);
+    scanner.getStaticWorkflowStates(getClass()).forEach(this::registerState);
```
gmokki, line 112: I thought the documentation said the states need to be registered. So do we need this scanner?
And if we do a scanner I would prefer that it would prefer a `static Collection<WorkflowState> getWorkflowStates()` method if such thing exists.
That would allow easier extending of workflow classes, where the extending class could filter out some parent states.
Maybe only if the method does not exist then we could register automatically?
Or can this be added later?

gmokki (commented):
I don't see any changes to nflow-examples? Do they require changes?

nflow-engine/src/main/java/io/nflow/engine/internal/workflow/WorkflowDefinitionScanner.java
efonsell: Added try-catch

efonsell:
bare-minimum and full-stack examples updated, full-stack-kotlin todo

gmokki (approved):

PR #434, opened 2021-02-17
add wait state to cron workflow

efonsell:


nflow-engine/src/main/java/io/nflow/engine/workflow/curated/CronWorkflow.java
```diff
@@ -97,7 +103,12 @@ protected CronWorkflow(String type, WorkflowSettings settings) {
    * <pre>
    * public NextAction doWork(StateExecution execution) {
    *   // do the work here
-   *   return NextAction.moveToState(schedule, "Work done");
+   *   if (waitForChildWorkflowsToFinish) {
```
eputtone, line 106: Wrong condition in the example? Better condition below?

```
if (rescheduleImmediately) {
  return NextAction.moveToState(schedule, "Work done");
} else {
  return NextAction.moveToStateAfter(waitForWorkToFinish, DateTime.now().plusHours(hoursToWait),
    "Waiting for work to finish");
}
```

efonsell: Fixed

eputtone (approved):
Looks good.

Pretty nice that completion of child workflow automatically wakes up the parent. I didn't know that. In the example there's no guard against child being still running. I've sometimes missed easy oneliner for checking that.

efonsell:
Easy oneliner added to `StateExecution` and used in the default implementation of `CronWorkflow.waitForWorkToFinishImpl`

eputtone:
Looks even better to my eye, still approved.

StateExecution.hasUnfinishedChildren() is exactly what I've missed occasionally. Though would StateExecution.hasUnfinishedChildWorkflows() be more consistent with the naming of other methods in StateExecution (?)

efonsell:
Renamed the method and moved the implementation to `WorkflowInstanceService` to make it generally available and to support checking children of any workflow instance. Added also some tests that revealed a bug in the query code (now fixed).

eputtone (approved):
Great, thanks! :+1:

PR #438, opened 2021-02-24
Fix webflux error handling

nsev:
Currently (at least) `NflowNotFoundException`s seem to get mapped to 500 instead of 404.

### DIFF
--- a/nflow-rest-api-common/src/main/java/io/nflow/rest/v1/ResourceBase.java
+++ b/nflow-rest-api-common/src/main/java/io/nflow/rest/v1/ResourceBase.java
@@ -1,6 +1,9 @@
 package io.nflow.rest.v1;
 
 import static io.nflow.engine.workflow.instance.WorkflowInstanceAction.WorkflowActionType.externalChange;
+import static java.net.HttpURLConnection.HTTP_BAD_REQUEST;
+import static java.net.HttpURLConnection.HTTP_INTERNAL_ERROR;
+import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
 import static java.util.Collections.sort;
 import static java.util.Collections.unmodifiableMap;
 import static java.util.stream.Collectors.toCollection;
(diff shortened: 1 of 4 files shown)


nflow-rest-api-common/src/main/java/io/nflow/rest/v1/ResourceBase.java
```diff
@@ -183,15 +183,22 @@ public ListWorkflowInstanceResponse fetchWorkflowInstance(final long id, final S
     return listWorkflowConverter.convert(instance, includes);
   }
 
+  protected int resolveExceptionHttpStatus(Throwable t) {
+    if (t instanceof IllegalArgumentException) {
+      return 400;
+    } else if(t instanceof NflowNotFoundException) {
```
efonsell, line 189:
```suggestion
    } else if (t instanceof NflowNotFoundException) {
```

nflow-rest-api-common/src/main/java/io/nflow/rest/v1/ResourceBase.java
```diff
@@ -183,15 +183,22 @@ public ListWorkflowInstanceResponse fetchWorkflowInstance(final long id, final S
...
+      return 404;
+    }
+
```
efonsell, line 192:
```suggestion
```

nflow-rest-api-common/src/main/java/io/nflow/rest/v1/ResourceBase.java
```diff
@@ -183,15 +183,22 @@ public ListWorkflowInstanceResponse fetchWorkflowInstance(final long id, final S
     return listWorkflowConverter.convert(instance, includes);
   }
 
+  protected int resolveExceptionHttpStatus(Throwable t) {
+    if (t instanceof IllegalArgumentException) {
+      return 400;
```
efonsell, line 188: Not added in this PR, but prefer using some existing constants for HTTP status codes. Same for 404 and 500.

efonsell (changes requested):

nflow-rest-api-common/src/main/java/io/nflow/rest/v1/ResourceBase.java
nsev: Sure - fixing.

efonsell:
Would it be possible to add some tests for this? I think the tests for netty are lacking in general, so not mandatory in this PR. But would be nice to have at least something...

efonsell (approved):

PR #441, opened 2021-03-09
Persist workflow instance query parameters to URL

eputtone:
Set the workflow instance query parameters to state, when search button is pressed. This again sets the parameters to URL, and reloads the search controller with the new query parameters, that are automatically used for making the actual query. When navigating away from the search page, browser back-button will work and the previous query result will be again rendered.

There's a corner case when first opening search page vs. searching without any criteria. The latter is handled by fake workflow definition type "all", which is ignored as a query parameter, but whose presence forces the search page to make the REST call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 **Highlights**
 - `nflow-explorer`
   - Sortable workflow definitions and instance search result table
+  - Persist workflow instance query parameters to URL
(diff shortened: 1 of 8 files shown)


efonsell (changes requested):
Just copy the PR description to the changelog :)

efonsell (approved):

PR #446, opened 2021-03-25
disable cache for oracle sequences

efonsell:


eputtone (approved):

nflow-engine/src/main/resources/scripts/db/update-7.2.5-x/oracle.update.ddl.sql
```diff
@@ -0,0 +1,17 @@
...
+/
+
+create or replace trigger nflow_workflow_action_insert
+  before insert on nflow_workflow_action
+  for each row
+declare
+begin
+  :new.id := nflow_workflow_action_id_seq.nextval;
+end;
+/
```
eputtone, line 17: Perhaps set nflow_workflow_action_id_seq to nflow_workflow_id_seq current value after replacing the trigger, so that there's minimal chance of colliding ids.

eputtone: Or action_seq=workflow_seq.nextval, as calling currentval requires calling nextval in the same session (if I remember my Oracle correctly from year 2004)

efonsell: Changed to recreate the sequence using `max(id) from nflow_workflow_action + 1000` **before** replacing the trigger. That way the trigger should generate unique ids immediately after it's replaced.

PR #466, opened 2021-12-20
Version bump

gmokki:


maven-version-rules.xml
```diff
@@ -36,6 +36,8 @@
         <rule groupId="org.hibernate.validator">
             <ignoreVersions>
```
jsyrjala, line 37: CHANGELOG.md needs an entry about these changes.

nflow-engine/src/main/resources/scripts/db/h2.create.ddl.sql
```diff
@@ -11,13 +11,13 @@ create table if not exists nflow_workflow (
...
-  next_activation timestamp,
-  external_next_activation timestamp,
+  next_activation timestamp with time zone,
+  external_next_activation timestamp with time zone,
   executor_id int,
   retries int not null default 0,
-  created timestamp not null default current_timestamp,
-  modified timestamp not null default current_timestamp,
-  started timestamp,
+  created timestamp with time zone not null default current_timestamp,
```
jsyrjala, line 18: Should there be any changes to scripts that change existing h2 databases?

maven-version-rules.xml
gmokki: Why? The maven-version-rules.xml is only for local developers so that they can run `mvn versions:update-properties` without getting versions that we do not work or that we for other reasons do no want (such as alpha/beta code).
Or would you like to have a section somewhere telling that which spring or servlet engines versions do work?

nflow-engine/src/main/resources/scripts/db/h2.create.ddl.sql
gmokki: hmm. Maybe not. But I'll add an entry to changes that says if you are using h2 as real database you must backup + recreate + restore backup. That is anyway the only h2 1.x to 2.x upgrade procedure.

efonsell (commented):
Should we drop Java 8 support or downgrade Jetty (10 -> 9) and HikariCP (5 -> 4)?

PR #475, opened 2022-01-16
Improve MAVEN build Performance

xiayingfeng:
Maven will run all tests in a single forked VM by default. This can be problematic if there are a lot of tests or some very memory-hungry ones. We can fork more test VM by setting `<fork>1.5C</fork>`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -212,6 +212,7 @@
             <parallel>classes</parallel>
             <perCoreThreadCount>true</perCoreThreadCount>
             <threadCount>1</threadCount>
+          	<forkCount>1.5C</forkCount>
           </configuration>
         </plugin>
         <plugin>


efonsell:
Seems to be causing random test failures in nflow-tests module, at least for me. The tests are probably not suitable for running concurrently. I would not merge this change, but ofc you can change it locally if it works for you.

gmokki:
maybe we could make the fork count a property with default 1C or just 1 (to run only single test). And that would allow others to override it from command line

pom.xml
```diff
@@ -212,6 +212,7 @@
             <parallel>classes</parallel>
             <perCoreThreadCount>true</perCoreThreadCount>
             <threadCount>1</threadCount>
+          	<forkCount>1.5C</forkCount>
```
efonsell, line 215:
```suggestion
          	<forkCount>1C</forkCount>
```

efonsell: Seems like those randomly failing tests have been improved a bit to enable more concurrency. 1.5C seemed to be too much for now, but 1C should work.

efonsell:
Closing this, as tests were improved and now we have `<forkCount>1C</forkCount>` in master

PR #491, opened 2022-02-02
Enable sqlserver tests on github actions

gmokki:


.github/workflows/build.yaml
```diff
@@ -27,4 +27,9 @@ jobs:
       - name: Create frontend package
         run: mvn -X -B -pl nflow-explorer install
       - name: Run the Maven test phase
-        run: mvn -X -B -P ${{ matrix.db }} test
+        run: mvn -X -B -P ${{ matrix.db }} -Dskip.npm -Dskip.bower -Dskip.grunt test
```
efonsell, line 30: We really want to skip all of these?

.github/workflows/build.yaml
```diff
@@ -27,4 +27,9 @@ jobs:
...
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v2
+        if: always() # always run even if the previous step fails
+        with:
+          report_paths: '**/target/surefire-reports/TEST-*.xml'
```
efonsell, line 35: newline missing at eof

scripts/setup-db-sqlserver.sh
```diff
@@ -1,14 +1,20 @@
...
-docker exec -t mssql /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P passWord1% -e -x -Q "create database nflow"
-docker exec -t mssql /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P passWord1% -e -x -d nflow -Q "create login [nflow] with password='nFlow42%', default_database=[nflow]"
-docker exec -t mssql /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P passWord1% -e -x -d nflow -Q "create user [nflow] for login [nflow] with default_schema=[nflow]"
-docker exec -t mssql /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P passWord1% -e -x -d nflow -Q "create schema nflow authorization nflow"
-docker exec -t mssql /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P passWord1% -e -x -d nflow -Q "grant connect to [nflow]"
-docker exec -t mssql /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P passWord1% -e -x -d nflow -Q "alter role db_datareader add member nflow"
-docker exec -t mssql /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P passWord1% -e -x -d nflow -Q "alter role db_datawriter add member nflow"
-docker exec -t mssql /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P passWord1% -e -x -d nflow -Q "grant all to [nflow]"
+VER=2019-latest
+if [[ "$1" == 11 ]]; then
```
efonsell, line 4: $1 could be assigned to a variable which is named in a way that documents the content

.github/workflows/build.yaml
gmokki: Yes. We already `install` the frontend package in the previous step. These only run `test`. These options thus this skips the double build of frontend.

PR #518, opened 2022-04-16
update deps

efonsell:


efonsell:
How about the remaining THROWS_METHOD_THROWS_CLAUSE_BASIC_EXCEPTION warnings?

PR #529, opened 2022-06-03
Bump latest versions of databases to test against: 

gmokki:
* db2: 11.5.0->11.5.7
* mariadb 10.7->10.8
* sqlserver 2019->2022


efonsell (approved):
lgtm

PR #535, opened 2022-06-04
Define maven.compiler.release=11

jsyrjala:
Ensures that we are compiling using JDK 11 version of APIs. Also allows building when using JDK 17.

https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html
https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html


efonsell (approved):
lgtm

PR #551, opened 2022-07-14
Try CodeQL analysis

gmokki:
Enables the github provided CodeQL analysis, results are visible in Security tab
Tunes the normal build to use newer versions of imported actions and adds some security settings to limit rights of the build.


efonsell (approved):
LGTM

PR #552, opened 2022-07-15
Optimize SQL for workflow recovery queries

gmokki:
Split recovery needed detection and listing of workflows needing recovery into separate transactions to fix slowdown with large amounts of workflow instances


nflow-engine/src/main/java/io/nflow/engine/internal/dao/ExecutorDao.java
```diff
@@ -164,27 +163,44 @@ private void updateWithPreparedStatement(String sql) {
...
+          var started = sqlVariants.getDateTime(rs, "started");
+          var active = sqlVariants.getDateTime(rs, "active");
+          var expires = sqlVariants.getDateTime(rs, "expires");
+          var stopped = sqlVariants.getDateTime(rs, "stopped");
+          var recovered = sqlVariants.getDateTime(rs, "recovered");
+          return new WorkflowExecutor(id, host, pid, executorGroup, started, active, expires, stopped, recovered);
+        });
   }
 
   public void markShutdown() {
```
gmokki, line 180: Isn't this supposed to be graceful shutdown that does not need any recovery? At least the old code did not recover any workflows from executors closed with this method.

nflow-engine/src/main/java/io/nflow/engine/internal/dao/ExecutorDao.java
```diff
@@ -164,27 +163,44 @@ private void updateWithPreparedStatement(String sql) {
...
+        "select id from nflow_executor where " + getExecutorGroupCondition() + " and id <> " + getExecutorId()
+            + " and recovered is null and " + sqlVariants.dateLtEqDiff("expires", "current_timestamp"),
+        (rs, rowNum) -> rs.getInt(1));
+  }
+
+  public void markRecovered(int recoveredExecutorId) {
+    try {
+      jdbc.update(
+          "update nflow_executor set recovered=current_timestamp where recovered is null and "
+              + sqlVariants.dateLtEqDiff("expires", "current_timestamp") + " and " + getExecutorGroupCondition() + " and id = ?",
```
gmokki, line 200: Do we need the expires or the executor group conditions here?

efonsell: If we assume that a node that has expired will never come back alive and start updating it's heartbeat, then they are not needed. Can we assume that?

nflow-engine/src/main/java/io/nflow/engine/internal/dao/ExecutorDao.java
efonsell: Yes, unless the graceful shutdown times out. Then we need to recover.

nflow-engine/src/main/java/io/nflow/engine/internal/dao/ExecutorDao.java
gmokki: ok

efonsell (approved):

PR #554, opened 2022-07-18
Improve graceful shutdown handling

gmokki:


nflow-engine/src/main/java/io/nflow/engine/internal/dao/ExecutorDao.java
```diff
@@ -177,10 +177,11 @@ public List<WorkflowExecutor> getExecutors() {
...
   }
 
-  public void markShutdown() {
+  public void markShutdown(boolean graceful) {
     try {
-      jdbc.update("update nflow_executor set expires=current_timestamp, stopped=current_timestamp, recovered=null where "
-          + getExecutorGroupCondition() + " and id = ?", getExecutorId());
+      var sql = graceful ? "set expires=current_timestamp, stopped=current_timestamp, recovered=current_timestamp" :
+              "set expires=" + sqlVariants.currentTimePlusSeconds(timeoutSeconds) + ", stopped=current_timestamp, recovered=null";
+      jdbc.update("update nflow_executor " + sql + " where id = ?", getExecutorId());
```
efonsell, line 184:
```suggestion
      var sql = "update nflow_executor set expires=current_timestamp, stopped=current_timestamp, recovered=? where id = ?";
      jdbc.update(sql, graceful ? "current_timestamp" : null, getExecutorId());
```

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@
 **Details**
 
 - `nflow-engine`
+  - Improve shutdown sequence.
+    - Workflows that have been acquired from database, but had not yet started executing, are now resumed immediately by another executor.
+    - During the last 5 seconds of shutdown timeout the workflows are interrupted so that they persist their state and can be resumed immediately.
```
efonsell, line 19:
```suggestion
    - Workflows that were acquired from the database but are not yet started executing can now be resumed immediately by another executor.
    - Executing workflows are interrupted 5 seconds before shutdown timeout so that they get a chance to persist their state to the database. This also allows other executors to immediately resume the processing of the successfully interrupted workflows.
```

nflow-engine/src/main/java/io/nflow/engine/internal/dao/ExecutorDao.java
gmokki: I would really prefer to give some extra delay to the in-progress workflows when a node is not shut down gracefully, which means that the workflow threads are still executing. Just to reduce the likelihood that the same instance is running on two executors concurrently.

efonsell: Yes, the comment was outdated.

nflow-engine/src/main/java/io/nflow/engine/internal/dao/ExecutorDao.java
```diff
@@ -177,10 +177,11 @@ public List<WorkflowExecutor> getExecutors() {
...
   }
 
-  public void markShutdown() {
+  public void markShutdown(boolean graceful) {
     try {
-      jdbc.update("update nflow_executor set expires=current_timestamp, stopped=current_timestamp, recovered=null where "
-          + getExecutorGroupCondition() + " and id = ?", getExecutorId());
+      var sql = graceful ? "set expires=current_timestamp, stopped=current_timestamp, recovered=current_timestamp" :
+              "set expires=" + sqlVariants.currentTimePlusSeconds(timeoutSeconds) + ", stopped=current_timestamp, recovered=null";
+      jdbc.update("update nflow_executor " + sql + " where id = " + getExecutorId());
```
efonsell, line 184: More reabable and uses prepared statement:
```suggestion
      var expires = graceful ? "current_timestamp" : sqlVariants.currentTimePlusSeconds(timeoutSeconds);
      jdbc.update("update nflow_executor set expires=?, stopped=current_timestamp, recovered=current_timestamp where id = ?",
          expires, getExecutorId());
```

gmokki: I don't think `sqlVariants.currentTimePlusSeconds` works as query parameter. All other uses are inlined.

efonsell: True. But the prepared statement argument still holds.

efonsell, line 184:
```suggestion
      var expires = graceful ? "current_timestamp" : sqlVariants.currentTimePlusSeconds(timeoutSeconds);
      jdbc.update("update nflow_executor set expires=" + expires
          + ", stopped=current_timestamp, recovered=current_timestamp where id = ?", getExecutorId());
```

PR #587, opened 2023-03-17
Add dedicated build for nflow-explorer-ng

jsyrjala:


.github/workflows/build-nflow-explorer-ng.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - '**/*.md'
+concurrency:
+  group: build-nflow-explorer-ng-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+permissions:
+  contents: read
+  checks: write
+  security-events: write
+jobs:
+  build-and-test-nflow-explorer-ng:
```
efonsell, line 17: I would put this job to the build.yaml with other jobs until duplicating the boilerplate around the job can be avoided.

.github/workflows/build-nflow-explorer.yaml
```diff
@@ -0,0 +1,35 @@
+name: Build and test nflow-explorer
```
efonsell, line 1: I think we should drop nflow-explorer, so this would not be needed

efonsell (changes requested):

.github/workflows/build-nflow-explorer-ng.yaml
jsyrjala: Doing it in separate .yaml makes it possible to run it in parallel with main build, thus making build faster.

efonsell: Different jobs in same file are ran in parallel, so having it in a separate file does not make build any faster.

.github/workflows/build-nflow-explorer.yaml
jsyrjala: dropped

.github/workflows/build-nflow-explorer-ng.yaml
jsyrjala: fixed

.github/workflows/build.yaml
```diff
@@ -36,6 +36,25 @@ jobs:
...
+          java-version: 11
+          distribution: 'temurin'
+          cache: 'maven'
+      - name: Run build and unit tests
+        run: mvn -T1C --no-transfer-progress --color=always -Dmaven.javadoc.skip=true -Dmaven.source.skip=true -am -pl nflow-explorer-ng test
+      - name: Publish JUnit test report
+        uses: scacap/action-surefire-report@v1
+        if: always()
+        with:
+          check_name: 'Unit Test Report for nflow-explorer-ng'
```
efonsell, line 57: This is failing as there are no results found, should this be removed?

jsyrjala: Removed

.github/workflows/build.yaml
```diff
@@ -31,11 +31,7 @@ jobs:
           cache: 'maven'
       - name: Run build and unit tests
         run: mvn -T1C --no-transfer-progress --color=always -Dmaven.javadoc.skip=true -Dmaven.source.skip=true -am -pl nflow-jetty,nflow-netty,nflow-metrics test
-      - name: Publish JUnit test report
-        uses: scacap/action-surefire-report@v1
-        if: always()
-        with:
-          check_name: 'Unit Test Report for Java ${{ matrix.java }}'
```
efonsell: Seems to be removed from java build, should be removed from explorer build?

efonsell (approved):

PR #589, opened 2023-03-17
Remove nflow-explorer

jsyrjala:
nflow-explorer has been obsoleted by nflow-explorer-ng.


efonsell:
Should we also rename nflow-explorer-ng as nflow-explorer? If not, update https://github.com/NitorCreations/nflow/blob/master/README.md

efonsell (approved):
Rename nflow-explorer-ng to nflow-explorer before releasing

Issue #595, opened 2023-05-09
nflow is not compatible with Spring Boot 3


cvasii:
nflow is not compatible with Spring Boot 3, because of the beans adnotated with javax.inject.Named (e.g. io.nflow.engine.internal.dao.MaintenanceDao). javax.inject.Named. is not available after upgrading to Spring Boot 3.

cvasii:
One quick fix could be replacing javax.inject.Named with spring's specific annotations.

efonsell:
Fixed in 9.0.0

PR #606, opened 2023-08-15
Bump to Spring Framework 6

tuomoa:
**Bump to Spring framework 6**
- Migrate from javax to jakarta
- Bump guice for jakarta support
- Bump jersey for jakarta support
- Bump apache cxf for jakarta support
- Bump dropwizard metrics and switch to jakarta servlets Bump jetty
- Switch jackson jaxrs to jakarta-rs
- Netty configuration needs Jackson2ObjectMapperBuilder to have findModulesViaServiceLoader enabled to find joda time module (it's not in wellknown configuration automatically)
  - Before spring 6 it was autoconfigured via wellknown modules
- Switch swagger jaxrs2 to jaxrs2-jakarta
- Bump hibernate validator for jakarta support
- Bump minimum JDK to 17
- Lot of other dependency updates also..

Implements #595


efonsell:
This requires a good changelog entry

nflow-engine/pom.xml
```diff
@@ -96,9 +96,13 @@
       <artifactId>joda-time</artifactId>
     </dependency>
     <dependency>
-      <groupId>javax.inject</groupId>
-      <artifactId>javax.inject</artifactId>
-      <version>1</version>
+      <groupId>jakarta.inject</groupId>
+      <artifactId>jakarta.inject-api</artifactId>
+      <version>2.0.1</version>
```
efonsell, line 101: versions should be defined in root pom

.github/workflows/build.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
     if: always()
     strategy:
       matrix:
-        java: [ '11', '17' ]
+        java: [ '17' ]
```
gmokki, line 21: I would add java 20 there for now (lets bump to 21 when next lts is released)

nflow-jetty/pom.xml
```diff
@@ -62,15 +62,15 @@
     </dependency>
     <dependency>
       <groupId>io.swagger.core.v3</groupId>
-      <artifactId>swagger-jaxrs2</artifactId>
+      <artifactId>swagger-jaxrs2-jakarta</artifactId>
```
efonsell, line 65: are the versions of the new deps defined in root pom? if not, they should be added.

maven-version-rules.xml
```diff
@@ -18,12 +18,12 @@
...
+        <rule groupId="jakarta.xml.bind">
             <ignoreVersions>
                 <ignoreVersion type="regex">.*-b[0-9]+.*</ignoreVersion>
             </ignoreVersions>
         </rule>
-        <rule groupId="org.glassfish" artifactId="javax.el">
+        <rule groupId="org.glassfish" artifactId="jakarta.el">
             <ignoreVersions>
                 <ignoreVersion type="regex">.*-b[0-9]+.*</ignoreVersion>
             </ignoreVersions>
```
gmokki, line 29: Please also change all locations in this file that have comment that never versions are held back because they require spring 6.x. And then run `mvn versions:display-property-updates` and bump pom.xml versions for those.

gmokki (changes requested):

maven-version-rules.xml
tuomoa: Sure will do.

tuomoa: Actually this was due to conflicting junit5 and junit platform commons versions. Working now.

gmokki (approved):

PR #609, opened 2023-08-18
Fix child workflow not loading custom instance content

tuomoa:
Fix child workflow not loading custom instance content


efonsell (approved):
LGTM

PR #615, opened 2023-08-22
Provide engine ObjectMapper via custom interface

tuomoa:
Provide `ObjectMapper` defined in `EngineConfiguration` via custom interface to avoid collision with the one in `RestConfiguration` and remove need for `Primary` annotation usage.


efonsell (approved):
LGTM

PR #641, opened 2024-05-03
Add Azure AD support to Explorer

eputtone:
Add Azure AD support to the new Explorer.

If `msalConfig` is defined in `config.js`, force the user through Azure AD redirect-based login into the defined application when opening the Explorer. In the example configuration, credentials are cached in `localStorage` so they are available in different browser tabs as it's relatively normal to open multiple workflows to different tabs.

Because API calls are in a centralized `service.ts`, we cannot easily use `msal-react` hooks to access the authentication. Hence, the `msalClient` created in when opening Explorer is somewhat dirtily piggy-backed in `config`-object to `service.ts` and used to acquire and enrich the API calls with `Bearer`-token.

I haven't tested what happens if all the tokens (including refresh token) are expired. Probably an unclean failure, from which the user can recover by refreshing the browser and starting over. I don't have an environment where API is protected by Azure AD, so I just checked that the `Authorization` token is generated and set correctly.

So it's somewhat experimental support, but better than nothing and can be improved based on feedback.

### DIFF
--- a/nflow-explorer/package-lock.json
+++ b/nflow-explorer/package-lock.json
@@ -8,6 +8,8 @@
       "name": "explorer",
       "version": "0.1.0",
       "dependencies": {
+        "@azure/msal-browser": "^3.13.0",
+        "@azure/msal-react": "^2.0.15",
         "@fontsource/roboto": "^4.5.8",
         "@material-ui/core": "^4.12.4",
         "@material-ui/icons": "^4.11.3",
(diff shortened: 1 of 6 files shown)


RealZimboGuy:
not sure if you would be interested, we have some code using spring security that can intercept and force a login to ldap ?

efonsell (approved):

ravening:
> not sure if you would be interested, we have some code using spring security that can intercept and force a login to ldap ?

@RealZimboGuy  can you share the code?

RealZimboGuy:
@ravening  here is the link to the repo

https://github.com/RealZimboGuy/nflow-spring-security-test

ravening:
> @ravening here is the link to the repo
> 
> https://github.com/RealZimboGuy/nflow-spring-security-test

@RealZimboGuy  thanks :)

PR #656, opened 2024-10-02
enhance graph to indicate workflow steps and show active state

RealZimboGuy:
can change the colours or what is wanted in the css file, the logic is done for the indications.

sample below, red text and underline for current state, then any state-state the arrow is also hi-lighted 

image

### DIFF
--- a/nflow-explorer/src/component/StateGraph.jsx
+++ b/nflow-explorer/src/component/StateGraph.jsx
@@ -20,22 +20,53 @@ import './StateGraph.scss';
 // - zoom and pan state graph
 
 // TODO move to service.ts?
-function createGraph(definition) {
+function createGraph(props) {
   const g = new dagreD3.graphlib.Graph().setGraph({});
-  const states = definition.states;
+  const states = props.definition.states;
   // create nodes
   for (let state of states) {
     g.setNode(state.id, {label: state.id, class: `node-${state.type}`});
     // Round the corners of the nodes
     const node = g.node(state.id);
     node.rx = node.ry = 5;
+
+    //if instance is provided, highlight with the .active
+    //
+    if (props.instance && props.instance.state === state.id) {
+      node.class += ' current-state';
+
+    }
+
+  }
+
+  function hasNavigatedState(fromState, toState, instance) {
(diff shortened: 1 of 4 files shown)


nflow-explorer/src/component/StateGraph.jsx
```diff
@@ -20,22 +20,53 @@ import './StateGraph.scss';
...
+    //if instance is provided, highlight with the .active
+    //
+    if (props.instance && props.instance.state === state.id) {
+      node.class += ' current-state';
+
+    }
+
+  }
+
+  function hasNavigatedState(fromState, toState, instance) {
```
eputtone, line 42: A more performant solution might traverse actions only once, and create a set of realized state transition pairs.

Then `hasNavigatedState()` would be just a lookup to this set, instead of iterating the actions for each transition in workflow definition.

But I can approve this solution as well, because loading `wordGenerator` (from nflow-tests) instance to the modified Explorer does not fail:
https://nflow.io/dev/index.html#/workflow/3

eputtone (commented):
Thanks for this!

Could the color for highlighting the workflow transitions and current state be e.g. blue instead of red? If it's not visible enough for you, maybe increase the size of the highlighting? I associate red with problems / alerts. 

Ignoring the color, I could merge this.

nflow-explorer/src/component/StateGraph.jsx
RealZimboGuy: thats a huge state diagram :)  ive changed it to blue as below 

image


i did think a bit on the performance side but its browser based and you end up with 
(number of workflow states) X(displayed actions (defaults to 100 based on system setting) 

overall not too many to worry about (i think)

eputtone: Yeah, the performance is sufficient, and there are more important things to polish in Explorer. For instance, we need to migrate from `dagre-d3` to something that is being maintained.

Thanks for changing the edge color to blue. I think I'd also slightly prefer blue in the current state highlight, though that's not important for me.  I'll merge the PR later today :+1:

PR #678, opened 2025-04-02
update selection to be a search select and display in alphabetic order

RealZimboGuy:
this change makes the selection lists searchable and display in alphabetic order 

image

image

image

### DIFF
--- a/nflow-explorer/src/component/Selection.tsx
+++ b/nflow-explorer/src/component/Selection.tsx
@@ -22,26 +22,29 @@ function Selection(props: {
 
   return (
     <FormControl style={{minWidth: 240}} variant="standard">
-            <Autocomplete
-                options={sortedItems}
-                value={props.selected}
-                onChange={(_, newValue: string | null) => props.onChange(newValue || '')}
-                renderInput={(params) => (
-                    <TextField
-                        {...params}
-                        label={props.label}
-                        variant="standard"
-                    />
-                )}
-                getOptionLabel={(option: string) => props.getSelectionLabel(option)}
-                renderOption={(itemProps, option: string) => (
-                    <li {...itemProps}>
+        <Autocomplete
+            options={sortedItems}
+            value={props.selected}
+            onChange={(_, newValue: string | null) => props.onChange(newValue || '')}
+            renderInput={(params) => (
+                <TextField
+                    {...params}
+                    label={props.label}
+                    variant="standard"
+                />
+            )}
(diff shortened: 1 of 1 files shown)


nflow-explorer/src/component/Selection.tsx
```diff
@@ -12,21 +12,36 @@ function Selection(props: {
...
+                renderInput={(params) => (
+                    <TextField
+                        {...params}
+                        label={props.label}
+                        variant="standard"
+                    />
+                )}
+                getOptionLabel={(option: string) => props.getSelectionLabel(option)}
+                renderOption={(itemProps, option: string) => (
+                    <li {...itemProps}>
```
eputtone, line 38: This causes an error (see browser developer console):
```
Selection.tsx:49 Warning: A props object containing a "key" prop is being spread into JSX:
  let props = {key: someKey, tabIndex: ..., role: ..., id: ..., onMouseMove: ..., onClick: ..., onTouchStart: ..., data-option-index: ..., aria-disabled: ..., aria-selected: ..., className: ..., children: ...};
  <li {...props} />
React keys must be passed directly to JSX without using spread:
  let props = {tabIndex: ..., role: ..., id: ..., onMouseMove: ..., onClick: ..., onTouchStart: ..., data-option-index: ..., aria-disabled: ..., aria-selected: ..., className: ..., children: ...};
  <li key={someKey} {...props} />
    at Autocomplete2 (http://localhost:3000/node_modules/.vite/deps/chunk-SJN37AS6.js?v=b5cab369:3847:17)
    at div
    at http://localhost:3000/node_modules/.vite/deps/chunk-BOFUXF5B.js?v=b5cab369:4301:50
    at FormControl2 (http://localhost:3000/node_modules/.vite/deps/chunk-TJPH5Y7O.js?v=b5cab369:7875:21)
    at Selection (http://localhost:3000/src/component/Selection.tsx:21:33)
  ...
```
I think it should be:
```
  renderOption={(itemProps, option: string) => {
      const { key, ...rest } = itemProps;
      return (
          <li key={key} {...rest}>
              {props.getSelectionLabel(option)}
          </li>
      );
  }}
```
So fix this error, and optionally indent `AutoComplete` block back one step (=indentation is not consistent in Explorer, but let's not make it worse). Then I'll merge this and make a new release as there's some other changes pending as well.

eputtone (approved):
Thanks for the contribution :rocket:

PR #685, opened 2025-06-24
add missing resource tag in pom file

RealZimboGuy:
demo spring boot pom file is missing the resource tag, this results in the default file being used which has a static url and port 3000 so the spring boot port defaults to 8080 and wont work

### DIFF
--- a/nflow-examples/spring-boot/full-stack/maven/pom.xml
+++ b/nflow-examples/spring-boot/full-stack/maven/pom.xml
@@ -45,6 +45,12 @@
 	</dependencies>
 
 	<build>
+		<resources>
+			<resource>
+				<directory>src/main/resources</directory>
+				<filtering>true</filtering>
+			</resource>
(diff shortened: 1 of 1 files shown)


nflow-examples/spring-boot/full-stack/maven/pom.xml
```diff
@@ -45,6 +45,12 @@
 	</dependencies>
 
 	<build>
+		<resources>
+			<resource>
+				<directory>src/main/resources</directory>
+				<filtering>true</filtering>
```
gmokki, line 51: why do we need filtering?
Otherwise looks good.
Also rebase to get a clean build.

RealZimboGuy: @gmokki  fixed on this recent commit "update java maven examples to use java 25 and spring boot 4 (#689)" closing PR

PR #698, opened 2026-05-17
format fixes, organize imports, add/remove warning suppressions

efonsell:


gmokki (approved):
Could you add an `.editorconfig` to the project with `ij_java_imports_layout` definition that forces all editors to use the same import order - can be done in separate PR too

efonsell:
> Could you add an `.editorconfig` to the project with `ij_java_imports_layout` definition that forces all editors to use the same import order - can be done in separate PR too

I will do a separate MR for that, and fix imports in all files then. This was mainly about fixing warnings.

PR #700, opened 2026-05-17
Migrate from jest to vitest

efonsell:


gmokki (approved):
No idea why we migrate, but LGTM

efonsell:
> No idea why we migrate, but LGTM

vitest is faster and goes well with vite

PR #701, opened 2026-05-20
Bump h2 to 2.4.x, which requires a workaround for a h2 bug 

gmokki:
that hopefully only materializes if database connection count goes to zero, and since no-one should be running h2 in production this should not matter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,10 @@
 - Dependency updates
   - spring 6.2.18
   - jetty 11.0.26
-  - apache cxf 4.2.0
+  - apache cxf 4.2.1
   - logback 1.5.32
   - jackson 2.21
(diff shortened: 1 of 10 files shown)


efonsell (approved):

efonsell:
After you get this working, consider updating h2 version in examples as well.

gmokki:
Now got it working. Actually found a way to fix the final bug a day ago. But took some time to improve the integration test framework so that the nflow-jetty/engine side warn/error log lines are propagated to the rest api caller.
Should make it much easier to fix any future problems in the nflow-tests package.

About h2 upgrade in examples. I think it might be a bit dangerous. Not sure if this upgrade in nflow even is sensible.
Any users of h2 for testing will notice the same problems and have to add a hack workaround.

PR #705, opened 2026-05-22
add .editorconfig to organize imports same way as eclipse does

efonsell:


efonsell:
@gmokki can you check if you get same order in IntelliJ Idea with this config?

gmokki (approved):

PR #707, opened 2026-05-22
Prepare for 11.0.0-SNAPSHOT

efonsell:


gmokki (approved):

gmokki:
not sure if it is a stupid idea, but I was thiinking we could do one final 10.0.2 release with latest deps for those that cannot jump to spring 7. then do identical release as 11 with spring 7 + apache license, but basically no new features.

efonsell:
> not sure if it is a stupid idea, but I was thiinking we could do one final 10.0.2 release with latest deps for those that cannot jump to spring 7. then do identical release as 11 with spring 7 + apache license, but basically no new features.

It's a good idea.

PR #709, opened 2026-05-22
Avoid worker into infinite retry loop with null name/value in setVariable

gmokki:
* And setVariable with null name throw NullPointer
* Make setVariable(String, String) with null value be ignored.

### DIFF
--- a/nflow-engine/src/test/java/io/nflow/engine/internal/dao/ExecutorDaoTest.java
+++ b/nflow-engine/src/test/java/io/nflow/engine/internal/dao/ExecutorDaoTest.java
@@ -90,15 +90,15 @@ public class ExecutorDaoTest extends BaseDaoTest {
   @Test
   public void markRecoveredSetsExpiredExecutorRecovered() {
     int crashedId = insertCrashedExecutor(dao.getExecutorGroup());
-    WorkflowExecutor executor = dao.getExecutors().get(0);
+    WorkflowExecutor executor = dao.getExecutors().stream().filter(e -> e.id == crashedId).findFirst().orElseThrow();
     assertThat(executor.recovered, is(nullValue()));
     assertThat(dao.getRecoverableExecutorIds(), contains(crashedId));
 
     dao.markRecovered(crashedId);
 
-    executor = dao.getExecutors().get(0);
+    executor = dao.getExecutors().stream().filter(e -> e.id == crashedId).findFirst().orElseThrow();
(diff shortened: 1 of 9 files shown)


gmokki:
When using `@StateVar` with `null` value the value is always ignored. So to be consistent with that the setVariable(String, String) will also just be ignored with null value. We could do the same with setVariable(String, Object) with null value, but I'm not sure if someone is using that api since it has worked before.

efonsell (approved):

nflow-engine/src/test/java/io/nflow/engine/internal/dao/ExecutorDaoTest.java
```diff
@@ -89,12 +89,12 @@ public void markShutdownSetsExecutorExpiredAndForcefullyStopped() {
 
   @Test
   public void markRecoveredSetsExpiredExecutorRecovered() {
-    insertCrashedExecutor(2, dao.getExecutorGroup());
+    int crashedId = insertCrashedExecutor(dao.getExecutorGroup());
     WorkflowExecutor executor = dao.getExecutors().get(0);
     assertThat(executor.recovered, is(nullValue()));
```
efonsell: if there's any non-crashed executor returned by this (in addition to the crashed one), this might fail. Maybe remove this and instead after checking that getRecoverableExecutorIds returns crashedId, find the crashed executor from the list of executors (instead of blindly using .get(0))

nflow-engine/src/test/java/io/nflow/engine/internal/dao/ExecutorDaoTest.java
efonsell, line 102:
```
  @Test
  public void markRecoveredSetsExpiredExecutorRecovered() {
    int crashedId = insertCrashedExecutor(dao.getExecutorGroup());
    WorkflowExecutor executor = dao.getExecutors().stream().filter(e -> e.id == crashedId).findFirst().orElseThrow();
    assertThat(executor.recovered, is(nullValue()));
    assertThat(dao.getRecoverableExecutorIds(), contains(crashedId));

    dao.markRecovered(crashedId);

    executor = dao.getExecutors().stream().filter(e -> e.id == crashedId).findFirst().orElseThrow();
    assertThat(executor.recovered, is(notNullValue()));
    assertThat(dao.getRecoverableExecutorIds(), not(contains(crashedId)));
  }
```

efonsell (approved):

PR #711, opened 2026-05-23
Add expectedState transition validation to workflow updates

efonsell:
- `nflow-engine`
  - Add optional state transition validation for workflow instance updates.
    - `WorkflowInstanceService.updateWorkflowInstance` now takes optional `expectedState` and returns `false` when transition from expected state to requested new state is not allowed by the workflow definition, or when the expected state does not match the instance state in the database.
    - Add `WorkflowDefinition.isAllowedStateTransition` helper for checking allowed transitions.
- `nflow-rest-api`
  - Add optional state transition validation for workflow instance updates.
    - Workflow instance update endpoints now take optional `expectedState` query parameter, pass it through to `WorkflowInstanceService.updateWorkflowInstance`, and return `HTTP 409 Conflict` transition from expected state to requested new state is not allowed by the workflow definition, or when the expected state does not match the instance state in the database.

### DIFF
--- a/nflow-engine/src/main/java/io/nflow/engine/service/WorkflowInstanceService.java
+++ b/nflow-engine/src/main/java/io/nflow/engine/service/WorkflowInstanceService.java
@@ -135,7 +135,8 @@ public class WorkflowInstanceService {
    * @param expectedState
    *          Optional expected current state for optimistic state transition check.
    * @param validationMode
-   *          Defines which state transitions are allowed, when expectedState is present.
+  *          Defines which state transitions are allowed, when expectedState is present. Must be
+  *          {@link StateTransitionValidationMode#doNotValidate} when expectedState is not present.
    * @return True if the update was successful, false otherwise.
    */
   @Transactional
@@ -146,6 +147,8 @@ public class WorkflowInstanceService {
     Assert.notNull(action, "Workflow instance action can not be null");
     Assert.notNull(expectedState, "Expected state can not be null");
     Assert.notNull(validationMode, "Validation mode can not be null");
+    Assert.isTrue(expectedState.isPresent() || validationMode == StateTransitionValidationMode.doNotValidate,
+        "Validation mode must be doNotValidate when expected state is not given");
     Assert.notNull(workflowDefinitionService, "workflowDefinitionService can not be null");
     try {
       WorkflowInstance.Builder builder = new WorkflowInstance.Builder(instance);
--- a/nflow-engine/src/main/java/io/nflow/engine/service/WorkflowInstanceService.java
+++ b/nflow-engine/src/main/java/io/nflow/engine/service/WorkflowInstanceService.java
@@ -102,17 +102,20 @@ public class WorkflowInstanceService {
 
   /**
    * Update the workflow instance in the database if it is currently not running, and insert the workflow instance action.
+   * If expectedState is present, update is allowed only when the current state matches it.
    * If the state of the instance is not null, the status of the instance is updated based on the new state.
    * If the state of the instance is null, neither state nor status are updated.
    * @param instance The instance to be updated.
    * @param action The action to be inserted.
+   * @param expectedState Optional expected current state for optimistic state transition check.
    * @return True if the update was successful, false otherwise.
    */
   @Transactional
   @SuppressFBWarnings(value = "WEM_WEAK_EXCEPTION_MESSAGING", justification = "NflowNotFoundException message is ok")
-  public boolean updateWorkflowInstance(WorkflowInstance instance, WorkflowInstanceAction action) {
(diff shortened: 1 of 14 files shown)


nflow-engine/src/main/java/io/nflow/engine/service/WorkflowInstanceService.java
```diff
@@ -102,17 +102,20 @@ public long insertWorkflowInstance(WorkflowInstance instance) {
...
    * If the state of the instance is null, neither state nor status are updated.
    * @param instance The instance to be updated.
    * @param action The action to be inserted.
+   * @param expectedState Optional expected current state for optimistic state transition check.
    * @return True if the update was successful, false otherwise.
    */
   @Transactional
   @SuppressFBWarnings(value = "WEM_WEAK_EXCEPTION_MESSAGING", justification = "NflowNotFoundException message is ok")
-  public boolean updateWorkflowInstance(WorkflowInstance instance, WorkflowInstanceAction action) {
+  public boolean updateWorkflowInstance(WorkflowInstance instance, WorkflowInstanceAction action, Optional<String> expectedState) {
```
gmokki: Lets add an old style wrapper that calls the new method to avoid forcing everyone to change their code (and not require the tests to be changed everywhere in nflow):

```java
public boolean updateWorkflowInstance(WorkflowInstance instance, WorkflowInstanceAction action) {
  return updateWorkflowInstance(instance, action, Optional.empty());
}
```

nflow-engine/src/test/java/io/nflow/engine/internal/dao/WorkflowInstanceDaoTest.java
```diff
@@ -498,10 +498,23 @@ public void updateNotRunningWorkflowInstanceDoesNotUpdateRunningInstance() {
...
 
+  @Test
+  public void updateNotRunningWorkflowInstanceDoesNotUpdateWhenExpectedStateDoesNotMatch() {
+    WorkflowInstance instance = constructWorkflowInstanceBuilder().build();
+    long id = dao.insertWorkflowInstance(instance);
+    WorkflowInstance modifiedInstance = new WorkflowInstance.Builder(instance).setId(id).setState("manualState").build();
+
+    boolean updated = dao.updateNotRunningWorkflowInstance(modifiedInstance, Optional.of("otherState"));
+
+    assertThat(updated, is(false));
```
gmokki, line 513: should we also try a working transition with the new argument?

nflow-engine/src/main/java/io/nflow/engine/service/WorkflowInstanceService.java
efonsell: I left it out intentionally to force users to decide if they want to check or not, but I added it back as deprecated instead.

nflow-engine/src/test/java/io/nflow/engine/internal/dao/WorkflowInstanceDaoTest.java
efonsell: Like updateWorkflowInstanceWorksWhenExpectedStateMatches?

gmokki: yes. but also on the Dao level that verifies the sql check allows updates to go through with valid value.

efonsell: for example `updateNotRunningWorkflowInstanceUpdatesStatusForNotRunningInstance` already covers that

nflow-engine/src/main/java/io/nflow/engine/service/WorkflowInstanceService.java
```diff
@@ -94,25 +95,57 @@ public long insertWorkflowInstance(WorkflowInstance instance) {
...
+   */
+  @Transactional
+  @SuppressFBWarnings(value = "WEM_WEAK_EXCEPTION_MESSAGING", justification = "NflowNotFoundException message is ok")
+  public boolean updateWorkflowInstance(WorkflowInstance instance, WorkflowInstanceAction action, Optional<String> expectedState,
+      StateTransitionValidationMode validationMode) {
     Assert.notNull(instance, "Workflow instance can not be null");
     Assert.notNull(action, "Workflow instance action can not be null");
+    Assert.notNull(expectedState, "Expected state can not be null");
+    Assert.notNull(validationMode, "Validation mode can not be null");
     Assert.notNull(workflowDefinitionService, "workflowDefinitionService can not be null");
```
gmokki, line 152: I would throw an IllegalArgumentException `if (validation != doNotValidate && expectedState == null)` because that combo does not currently work. Better catch that problem early.

efonsell: Fixed

gmokki (approved):

Issue #674, opened 2025-03-06
State change lost on object variable if not set read only


ptorkko:
In this example the state change is lost:

    public record ComplexObject(...) {}

    public NextAction myFancyState(StateExecution e, @StateVar("complex") ComplexObject o) {
        ...
        e.setVariable("complex", new ComplexObject(....));
        ...
    }

`ObjectStringMapper#storeArguments` overwrites the change made via `StateExecution#setVariable` because the `@StateVar` annotation does not have `readOnly = true`. Doesn't seem to apply to primitive or "simple" types as they get interpreted as read only due to their types.

Debatable if bug or feature, but definitely confusing and error prone.

Applies to version 10 at least.

efonsell:
Will be fixed in 11.0.0 (https://github.com/NitorCreations/nflow/pull/715)

efonsell:
`11.0.0-SNAPSHOT` now available in Maven Central snapshots repository if you want to give it a try

efonsell:
11.0.0 is release, closing this.

PR #714, opened 2026-05-24
Potentially breaking change: setVariable takes precedence over StateVar annotation

efonsell:
Do not merge before Spring 7 MR has been merged.

`StateExecution.setVariable(...)` now takes precedence over `@StateVar` argument write-back when both target the same state variable in one state execution

- `ObjectStringMapper.storeArguments(...)` now skips persisting state variables that were explicitly updated via `StateExecution.setVariable(...)`
- Log a warning when this conflict is detected and argument write-back is skipped
- Previous behavior: `@StateVar` argument write-back could override a value set via `StateExecution.setVariable(...)`


gmokki:
Good catch that the current implementation might be unexpected or at least undocumented.
But what would be a good reason to ever use setVariable if @StateVar is already declared for the step method?

But what would you think of the opposite? Throw an IllegalStateException on setVariable if the StateVar for the same variable name is declared in the method?

efonsell:
> Good catch that the current implementation might be unexpected or at least undocumented.

Kudos to @ptorkko for reporting https://github.com/NitorCreations/nflow/issues/674

> But what would be a good reason to ever use setVariable if @StateVar is already declared for the step method?
> But what would you think of the opposite? Throw an IllegalStateException on setVariable if the StateVar for the same variable name is declared in the method?

Yeah, that was one option. Maybe it's better. I could make it configurable if needed: either throw an error or log a warning when setVariable is called for StateVar.

efonsell:
Opened https://github.com/NitorCreations/nflow/pull/715 to replace this

PR #715, opened 2026-05-24
POTENTIALLY BREAKING CHANGE: Make setVariable throw error when called for @StateVar

efonsell:
`StateExecution.setVariable(...)` now throws error if called for any `@StateVar` state variables.
  - Either stop calling `StateExecution.setVariable(...)` or remove `@StateVar` state variable.
  - Previous behavior: `StateExecution.setVariable(...)` was silently ignored.

### DIFF
--- a/nflow-engine/src/main/java/io/nflow/engine/internal/workflow/StateExecutionImpl.java
+++ b/nflow-engine/src/main/java/io/nflow/engine/internal/workflow/StateExecutionImpl.java
@@ -10,12 +10,12 @@ import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.springframework.util.Assert;
 
-import jakarta.annotation.Nonnull;
 import io.nflow.engine.internal.dao.WorkflowInstanceDao;
 import io.nflow.engine.model.ModelObject;
 import io.nflow.engine.service.WorkflowInstanceService;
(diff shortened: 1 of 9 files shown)


nflow-engine/src/main/java/io/nflow/engine/internal/workflow/StateExecutionImpl.java
```diff
@@ -135,6 +139,9 @@ public void setVariable(@Nonnull String name, String value) {
       return;
     }
     workflowDao.checkStateVariableValueLength(name, value);
+    WorkflowStateMethod method = definition.getMethod(instance.state);
+    Assert.isTrue(method == null || Stream.of(method.params).noneMatch(p -> p.key.equals(name)),
```
gmokki: maybe move the matching logic to `WorkflowStateMethod.hasParameter(String name)`

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/StateExecution.java
```diff
@@ -193,4 +193,9 @@ public interface StateExecution {
    * @return True if unfinished child workflow instances are found, false otherwise.
    */
   boolean hasUnfinishedChildWorkflows();
+
+  /**
+   * For internal usage (ObjectStringMapper) only.
+   */
+  void setVariableInternal(String name, String value);
```
gmokki: Annoying. Could we internally refer to the StateExecutionImpl in the code that needs this so that we do not have to pollute the public interface?

nflow-engine/src/main/java/io/nflow/engine/internal/workflow/StateExecutionImpl.java
efonsell: moved

nflow-engine/src/main/java/io/nflow/engine/workflow/definition/StateExecution.java
efonsell: You can give it a try, I got circular reference which was not trivial to resolve.

efonsell: Hmm let's see if this works

efonsell: Is this better?

gmokki (approved):

PR #720, opened 2026-06-08
Do recovery in smaller batches

gmokki:


nflow-engine/src/main/resources/nflow-engine.properties
```diff
@@ -65,6 +65,7 @@ nflow.db.max_pool_size=4
 nflow.db.idle_timeout_seconds=600
 nflow.db.create_on_startup=true
 nflow.db.disable_batch_updates=false
+nflow.db.max_in_parameters=1000
```
efonsell, line 68: rename to `max_sql_in_parameters` to give a bit more context what this is?

